Issue #7, opened 2023-01-29
Lidar Object Detection Node


Gongsta:
Port over the lidar object detection node for https://git.uwaterloo.ca/WATonomous/wato_monorepo/-/tree/perp/add-pp/src/lidar_cuda into our lidar_object_detection node. Original repo is here: https://github.com/open-mmlab/mmdetection3d

Obstacle.msg: https://git.uwaterloo.ca/WATonomous/wato_monorepo/-/blob/develop/src/ros_msgs/common_msgs/msg/Obstacle.msg

- Read in a lidar scan from the `/lidar` topic
- Run inference with pointpillars
- output 3D bounding box predictions to `/detections_3d`

Another person will also need to help out with creating visualization tools, annotating the image directly and visualizing that.

PR #14, opened 2023-03-08
Porting over lidar node from wato_monorepo to wato_monorepo_v2

isaacwen:
Converted `lidar_cuda` node that was originally from wato_monorepo from ROS1 to ROS2, and made changes according to specs outlined in #7.


Edwardius:
@leungjch was wondering if this branch is still being used for lidar detection

leungjch:
@Edwardius no, will make a new PR which has some updates from this branch (using humble instead of foxy). thanks @isaacwen for your contributions!

PR #21, opened 2023-04-22
Adding the Radar RViz node in the Radar ROS2 Pipeline

rijin113:
Adding a ROS2 node that allows radar detections to be visualized in RViz.

RViz_Point_clouds

To run, edit `dev_config.local.sh` and add `ACTIVE_PROFILES="perception vis_tools"`, and run `watod2 build`; `watod2 up`
Then, enter the vnc container and start RViz by running `rviz2`. Open RViz in vnc viewer by connecting to the port. The port can be found using `watod2 -lp`. Once RViz is setup, enter the radar_rviz container and run `ros2 run radar_rviz radar_rviz_node`. 

Note: One way to test the node is to include a sample radar detection in the main function of the radar_rviz_node.cpp file.


Edwardius:
Is the radar visualization hardware intensive? If not, I would rather have it run automatically on startup of the radar node

Edwardius:
Confirm that this visualization works on foxglove

rijin113:
- Implemented this on foxglove and it seems to work in the same way with a bunch of random radar points that I made up. 
- When the container starts up, the node automatically launches and starts to publish data to topics immediately for foxglove to visualize. Is this what you mean?

image

Edwardius:
> - Implemented this on foxglove and it seems to work in the same way with a bunch of random radar points that I made up. 
> 
> - When the container starts up, the node automatically launches and starts to publish data to topics immediately for foxglove to visualize. Is this what you mean?
> 
> 
> 
> image
> 
> 

Yup, looks good. Rebase and we can merge.

PR #27, opened 2023-05-22
added foxglove container and documentation

OLeather:


Edwardius:
@OLeather did a rebase, and made the foxglove service act the same as the one in the asd_training repo. Double check, and then I can merge

Edwardius:
the foxglove docs might be out of date, added the port forwarding fix @VishGit1234 did for the training repo.

PR #33, opened 2023-09-13
working docker cleanup

Edwardius:
This change cleans up the fixuid commands from the dockerfiles, and also changes the watod2 command back to watod.

The fixuid commands should be treated as a black box unless if permission issues exist within the container. Please make sure that your repositories and files interact within /home/docker or /tmp/


Bert-Sun (approved):
lgtm

Edwardius:
> lgtm, personally id separate the uid blackboxing and watod2->watod renaming into 2 separate commits in case we need to roll one back but realistically i dont think there should be any issues with either

yeah I got lazy, we'll start taking commits seriously... soon

PR #34, opened 2023-09-20
Add camera YOLOv8 object detection node

leungjch:
Adds a ROS2 node for yolov8 camera object detection. Currently supports subscribing to one camera. Publishes an `ObstacleList` msg with the detections as well as an `SensorImage` message with the annotated bounding boxes and classes detected.

Tested on Nuscenes / data_stream node (running through foxglove)

https://github.com/WATonomous/wato_monorepo/assets/28817028/108b2ad5-4d31-44a3-b936-95d7d3c7a467


Gongsta:
See https://github.com/WATonomous/wato_monorepo/pull/48

PR #36, opened 2023-09-23
Removed fixuid from the radar object detection dockerfile

rijin113:
In PR https://github.com/WATonomous/wato_monorepo/pull/21, it was noticed that the radar_object_detection.dockerfile was not fully up to update. This PR is to update the dockerfile.


Gongsta:
Outdated PR. This is taken care of in the base wato docker image now.

PR #46, opened 2024-01-09
[fix] unspecify ros_distro for rosdep

Edwardius:
rosdep was using noetic rosdistro


Edwardius:
Forcing merge because it's fairly trivial.

PR #75, opened 2024-02-10
Working ros drivers for multiple flir cameras and velodyne lidar

Gongsta:


brianzheng206:
frames publish at a constant 24-28hz with around 4-7% drops

PR #79, opened 2024-02-21
Upgrade to medium size model

Gongsta:
image

Better accuracy. We can either use YOLOv8m or YOLOv8l (YOLOv8x is overkill). Settling for medium as it seems a good compromise.

Also increased default model resolution


Edwardius (approved):
lgtm, can you add some info on this? total vram usage increase, timing, and accuracy increase (opt)

Gongsta:
vram is ~2x more used per model

PR #80, opened 2024-02-21
Add traffic light launch file

Gongsta:
Add a traffic light launch file, which uses a fine-tuned YOLOv8 model that is mounted and launched through the `CameraObjectDetection` node.


Gongsta:
change in coordinate frames (back into the original image) will be fixed in another pr

src/perception/camera_object_detection/camera_object_detection/yolov8_detection.py
```diff
@@ -90,7 +101,51 @@ def __init__(self):
...
+                    bbox[2] * width_scale,
+                    bbox[3] * width_scale]
+        elif self.crop_mode == "CenterCrop":
+            translation = (self.orig_image_width / height_scale - self.image_size) / 2
+            return [(bbox[0] + translation) * height_scale,
+                    bbox[1] * height_scale,
+                    bbox[2] * height_scale,
+                    bbox[3] * height_scale]
+
+    def crop_and_convert_to_tensor(self, cv_image):
```
leungjch, line 148: maybe we can move these functions into a shared `perception_utils` folder, if these will be used by other nodes like traffic light detection

Lane detection and semantic segmentation will probably use these but are in C++ so we will need to reimplement them in C++ as well

leungjch (approved):

leungjch:
lgtm, seems like theres merge conflicts

modules/docker-compose.perception.yaml
```diff
@@ -42,16 +43,6 @@ services:
       target: deploy
     image: "${PERCEPTION_LIDAR_OBJECT_DETECTION_IMAGE}:${TAG}"
     command: /bin/bash -c "ros2 launch lidar_object_detection lidar_object_detection.launch.py"
-  traffic_light_detection:
-    build:
-      context: ..
-      dockerfile: docker/perception/traffic_light_detection/traffic_light_detection.Dockerfile
-      cache_from:
-        - "${PERCEPTION_TRAFFIC_LIGHT_DETECTION_IMAGE}:${TAG}"
-        - "${PERCEPTION_TRAFFIC_LIGHT_DETECTION_IMAGE}:main"
```
Edwardius, line 51: same as https://github.com/WATonomous/wato_monorepo/pull/83 , please remove all remnants of `PERCEPTION_TRAFFIC_LIGHT_DETECTION_IMAGE` and its derivatives in the repo

src/perception/camera_object_detection/camera_object_detection/yolov8_detection.py
Edwardius: this might make docker building a bit tricky, plz keep utils within a single container (ie. object detection utils, not a perception-wide utils)

Edwardius: price to pay for modularity

Edwardius (approved):
Lgtm, just address comments

Edwardius:
also add description of pr

src/perception/camera_object_detection/camera_object_detection/yolov8_detection.py
Gongsta: IMO Don't need to prematurely modularize, code is ~300 lines right now, once the codebase grows we can split it into different files. Hurts readability to have the logic spread out everywhere

Edwardius (approved):

PR #81, opened 2024-02-23
Converting -dev to a Watod-config parameter

Edwardius:
Two changes:
- Changes -dev to a new MODE_OF_OPERATION in the watod-config
- Allows watod-config.local.sh which is a local override that you can create that watod reads

### DIFF
--- a/watod
+++ b/watod
@@ -75,8 +75,10 @@ if [[ $# -gt 0 ]]; then
 	COMPOSE_CMD="${COMPOSE_CMD} $@"
 fi
 
-# Allow for local overrides of any of the below parameters
-if [ -f "$MONO_DIR/watod-config.sh" ]; then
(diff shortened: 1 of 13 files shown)


watod
```diff
@@ -88,55 +75,26 @@ if [[ $# -gt 0 ]]; then
...
-            echo "		ssh$LOCAL_FORWARD $USER@$HOSTNAME"
-            echo "	on your local machine attach to VNC at localhost:${VNC_PORT}"
-            echo
-        fi
-    done
-    
-    echo "========================================================================="
-    echo "To forward all ports locally:"
-    echo "ssh $LOCAL_FORWARD_ALL $USER@$HOSTNAME"
-    echo "========================================================================="
```
Edwardius, line 139: I removed this because its not used anymore, and vnc should be deprecated eventually

Gongsta (approved):

PR #83, opened 2024-03-01
Add traffic sign launch file

Miniapple8888:
* Add traffic signs launch file to run object detection for traffic signs. 
* Remove prior references to Perception traffic signs image (not needed anymore, it will run on camera object detection node). 
* Modify camera objection detection node to include CenterCrop and save annotated images (changes also incorporated in traffic light launch PR #80).


Edwardius:
plz add description of pr

modules/docker-compose.perception.yaml
```diff
@@ -53,17 +55,6 @@ services:
     image: "${PERCEPTION_TRAFFIC_LIGHT_DETECTION_IMAGE}:${TAG}"
     command: /bin/bash -c "ros2 launch traffic_light_detection traffic_light_detection.launch.py"
 
-  traffic_sign_detection:
-    build:
-      context: ..
-      dockerfile: docker/perception/traffic_sign_detection/traffic_sign_detection.Dockerfile
-      cache_from:
-        - "${PERCEPTION_TRAFFIC_SIGN_DETECTION_IMAGE}:${TAG}"
```
Edwardius, line 61: please also remove all remnants of PERCEPTION_TRAFFIC_SIGN_DETECTION_IMAGE and its derivatives in the repo

modules/docker-compose.perception.yaml
```diff
@@ -30,7 +31,8 @@ services:
               capabilities: [ gpu ]
     command: /bin/bash -c "ros2 launch camera_object_detection eve_launch.py"
```
Edwardius, line 32: on startup, this means that we only launch the generic object detectors for Eve. Do we also want to launch the traffic light detector? If so, I would combine the launch files of eve_launch and traffic_signs_launch

src/perception/camera_object_detection/config/traffic_signs_config.yaml
```diff
@@ -0,0 +1,9 @@
+traffic_signs_node:
+  ros__parameters:
+    camera_topic: /camera/left/image_color
+    publish_vis_topic: /traffic_signs_viz
+    publish_detection_topic: /traffic_signs
+    model_path: /perception_models/traffic_signs_v1.pt
+    crop_mode: CenterCrop
+    image_size: 1024
```
Edwardius, line 8: @Gongsta I was mislead by this parameter because it actually doesn't mean the size of the input image from the camera topic, its actually the size in which the image is reshaped before going into the object detector.

How camera-agnostic is Yolov8?

Edwardius (approved):
LGTM, please address some small changes. @Gongsta I got a question for you here as well

src/perception/camera_object_detection/config/traffic_signs_config.yaml
Miniapple8888: The reason for `CenterCrop` and specifying `image_size` is so that the predictions can be more accurate. The `image_size` was initially, by default, set to `640` which removes some detail from the image which reduces prediction accuracy since objects being detected like traffic lights and traffic signs are really small and get blurred out by the resizing (done by Letterbox too). `CenterCrop` crops the image so we don't lose too much detail when resizing.

I'm not sure by what you mean by camera-agnostic. The dataset the Yolov8 model is trained on is a collection of images from left camera, right camera, and center camera (that has fisheye lens). We're planning on removing images from center camera since it might affect how signs are detected in general. According to Steven, the left camera has the most high resolution pictures.

Edwardius: nw about this @Miniapple8888 , i was asking cuz I feel like the ros input image size should be specified in the config rather than determined on the fly. But I guess the current way is fine too as it reduces the number params we have to always reset and whatnot

modules/docker-compose.perception.yaml
Edwardius: @Gongsta @Miniapple8888 thoughts on this before we merge?

Miniapple8888: i think we should combine them (and do the same for traffic light detection) since both are using camera object detection anyways. I'll make that change in the code.

modules/docker-compose.perception.yaml
```diff
@@ -30,7 +31,8 @@ services:
...
     volumes:
-      - /mnt/wato-drive2/perception_models/yolov8m.pt:/perception_models/yolov8m.pt
+      - /mnt/wato-drive2/perception_models/yolov8s.pt:/perception_models/yolov8s.pt
```
Miniapple8888, line 34: Note for myself: change to `yolov8s.pt`

PR #86, opened 2024-03-05
Skip sim ci

VishGit1234:
band-aid fix for sim breaking ci (basically just skip some steps for sim services)


Edwardius (changes requested):
Instead, can u skip sim when the matrix is being created? this is done in docker_context.sh

Edwardius (approved):

PR #96, opened 2024-03-21
Add depth estimation ROS 2 node starter code

leungjch:
Starter code for C++ ROS2 node for depth estimation


Edwardius (approved):
LGTM, just make sure it runs in CI and whatnot

PR #99, opened 2024-03-22
Hepromark/tilt

hepromark:
Watod Tilt - WIP

Background:
- Tilt is used for development work with containers, an industry version of our internal watod tool. 

Goals:
- The container management will be implemented using Tilt, with watod as the user-facing CLI remaining the same. 
- Eliminate the need for dev_overrides folder using Tilt's ability to edit docker-compose files during runtime.


watod
```diff
@@ -24,14 +24,39 @@ function usage {
     exit 0
 }
 
+function tilt_up {
+    cd modules
+    echo "Runing tilt up with" $COMPOSE_MODULES
```
Edwardius, line 29: running

Edwardius:
@hepromark This still WIP?

Edwardius:
tilt not used

PR #173, opened 2025-01-27
copied package and faking odom values

hassaan141:
Faking odom values for the odometery node, these values are based of values that we might get in a car, so varaible speed with variable turn angles


src/world_modeling/odom/src/dummyData.cpp
VishGit1234, line 1: The message types need to be changed here from Int32 to Float64. Other than that it looks fine, since this is just dummy data you don't need to split this up into header and source files or anything.

VishGit1234 (changes requested):

VishGit1234 (approved):

VishGit1234 (approved):
Cool!

PR #187, opened 2025-04-04
Depth estimation perf

shendrew:
End to end performance optimization for image inference and point cloud serialization (3fps -> 20fps):
 - resize image input to 518x518 for depth estimation (recommended), then interpolate back to HD
 - refactor object creation in during serialization to reduce memory accesses
 - remove unused point cloud information (ie. color)
 - also refactor docker image to cache DepthAnything pip dependencies earlier for faster rebuild times

Suggest to be merged before integrating depth inference with TensorRT by @conjeevaram

### DIFF
--- a/src/perception/depth_estimation/depth_estimation/depth_estimation.py
+++ b/src/perception/depth_estimation/depth_estimation/depth_estimation.py
@@ -25,18 +25,24 @@ class DepthAnything(Node):
         self.declare_parameter("publish_depth_pcl_topic", "/camera/right/depth_plc")
         self.declare_parameter("small_model_path", "/perception_models/depth_anything_v2_small.pth")
         self.declare_parameter("large_model_path", "/perception_models/depth_anything_v2_large.pth")
+        self.declare_parameter("focal_length_x", 470.4)
+        self.declare_parameter("focal_length_y", 470.4)
+        self.declare_parameter("min_depth_offset", 0)
+        self.declare_parameter("max_depth_cuttoff", 79)
 
         self.camera_topic = self.get_parameter("camera_topic").value
         self.publish_depth_img_topic = self.get_parameter("publish_depth_img_topic").value
         self.publish_depth_pcl_topic = self.get_parameter("publish_depth_pcl_topic").value
         self.small_model = self.get_parameter("small_model_path").value
         self.large_model = self.get_parameter("large_model_path").value
+        self.focal_length_x = self.get_parameter("focal_length_x").value
+        self.focal_length_y = self.get_parameter("focal_length_y").value
+        self.min_depth_offset = self.get_parameter("min_depth_offset").value
+        self.max_depth_cuttoff = self.get_parameter("max_depth_cuttoff").value
 
         self.encoder_choices = ['vits', 'vitb', 'vitl', 'vitg']
         self.encoder = self.encoder_choices[0]
         self.max_depth = 80  # in meters, I believe
-        self.focal_length_x = 470.4
-        self.focal_length_y = 470.4
         self.outdir = "/runs/"
         self.img_id = 0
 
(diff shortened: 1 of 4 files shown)


src/perception/depth_estimation/depth_estimation/depth_estimation.py
```diff
@@ -12,7 +12,7 @@
 from sensor_msgs.msg import CompressedImage, Image, PointCloud2, PointField
 from rclpy.node import Node
 
-from depth_anything_v2.dpt import DepthAnythingV2
```
FowlFarmer, line 15: wait what is the difference here?

src/perception/depth_estimation/depth_estimation/depth_estimation.py
```diff
@@ -71,7 +71,7 @@ def __init__(self):
             CompressedImage,
             self.camera_topic,
             self.image_callback,
-            10
+            60
```
FowlFarmer, line 74: if i remember correctly, the queue gets rid of the older msg if the queue is full and a new msg arrives. If we want the newest message always to be processed by the model, couldn't we reduce this queue size?

FowlFarmer (approved):

src/perception/depth_estimation/depth_estimation/depth_estimation.py
shendrew: I see, was using it to test. Should be fixed now

src/perception/depth_estimation/depth_estimation/depth_estimation.py
shendrew: The previous one outputs values from 0-1, metric tries to output in meters

lucasreljic:
The branch needs to be updated with main, otherwise there will be conflicts

src/perception/depth_estimation/depth_estimation/depth_estimation.py
```diff
@@ -85,24 +86,37 @@ def __init__(self):
...
-        cloud_msg.row_step = cloud_msg.point_step * points.shape[0]
-        cloud_msg.is_dense = True
-        # Convert the points array to a bytes object.
-        cloud_msg.data = points.astype(np.float32).tobytes()
-        return cloud_msg
+        # Create the output directory if it doesn't exist
+        os.makedirs(self.outdir, exist_ok=True)
+
+        # test throughput by spamming publishes
+        # for i in range(20):
```
lucasreljic, line 93: Remove unused code

src/perception/depth_estimation/depth_estimation/depth_estimation.py
```diff
@@ -33,7 +33,7 @@ def __init__(self):
         self.large_model = self.get_parameter("large_model_path").value
 
         self.encoder_choices = ['vits', 'vitb', 'vitl', 'vitg']
-        self.encoder = self.encoder_choices[2]
+        self.encoder = self.encoder_choices[0]
         self.max_depth = 80  # in meters, I believe
         self.focal_length_x = 470.4
```
lucasreljic, line 38: Focal length should be in the config.yaml file

src/perception/depth_estimation/depth_estimation/depth_estimation.py
```diff
@@ -115,45 +129,47 @@ def image_callback(self, msg):
...
         # === generate point cloud ===
-        t_pcl_start = time.time()
+        t_pcl_numpy_start = time.time()
         x, y = np.meshgrid(np.arange(width), np.arange(height))
         x = (x - width / 2) / self.focal_length_x
         y = (y - height / 2) / self.focal_length_y
-        z = np.array(resized_pred)
-        points = np.stack((np.multiply(x, z), np.multiply(y, z), z), axis=-1).reshape(-1, 3)
-        colors = np.array(color_image).reshape(-1, 3) / 255.0
+        z = metric_pred + 0        # constant should be adjusted to fit ground truth
```
lucasreljic, line 130: If this needs to be adjusted it should also be in the config file

src/perception/depth_estimation/depth_estimation/depth_estimation.py
shendrew: updated, suggest to keep output directory for visualization

lucasreljic (changes requested):
Looks good, make sure there is no hard coded values that might need to be configured later

src/perception/depth_estimation/depth_estimation/depth_estimation.py
shendrew: updated

lucasreljic:
What is causing the long build time >10 minutes for the container? It's not that big of an issue though we can just push an image to the registry

lucasreljic (approved):
LGTM!

shendrew:
> What is causing the long build time >10 minutes for the container? It's not that big of an issue though we can just push an image to the registry

Hmm not sure.. on slurm it takes ~3 min for a cold build. The depthanything dependencies are probably the longest

PR #214, opened 2025-07-26
[World Modeling] Conglomerated World Modeling Dockerfiles

JStephenHuang:
### 📑  Description
Made a global world modeling dockerfile to run world modeling on one image. 

No demo, but I do have some notes...

### ✅  Checklist

### 🔗  Related Issues / PRs
No related issues

### 📝  Notes for reviewers
After the #205 commit, there was some build errors notably in the hd map, localization and occupancy_segmentation nodes. The reoccurring build error was the added `const` in some of the parameters of methods. Those parameter's value could not be kept constant due to the logic under the methods. After the fixes...

the entirety of wm does run now on one container. However, needs approval in terms of the cleanness of the code specifically in the `world_modeling_bringup` and the dockerfile.

Also the `watod-setup-env.sh` script had an error for multi modules build.


JStephenHuang:
dude what the... how do i fix this conflict...

docker/world_modeling/occupancy_segmentation/occupancy_segmentation.Dockerfile
Edwardius, line 1: nice!!

docker/world_modeling/world_modeling.Dockerfile
```diff
@@ -1,32 +1,43 @@
...
 FROM ${BASE_IMAGE} AS dependencies
 
-# INSTALL DEPENDENCIES HERE BEFORE THE ROSDEP
-# Only do this as a last resort. Utilize ROSDEP first
+# Download maps and install apt deps
+ENV MAPS_DIR="${AMENT_WS}/etc/maps/"
+RUN apt-get update && \
+    git clone https://github.com/WATonomous/map_data.git --depth 1 "${MAPS_DIR}" && \
+    chmod -R 755 "${MAPS_DIR}" && \
+    rm -rf /var/lib/apt/lists/*
```
Edwardius: design level, we should make a file location in the monorepo (assets) containing large file structures like model weights, maps, etc. There should then exist some scripting in watod to import those files via something like curl or something.

out of scope for this PR, but could you make a separate pr that creates an assets dir? and an addition to watod to automatically download world modeling assets (option to specify the subgroup to download assets for)

src/world_modeling/hd_map/src/hd_map_manager.cpp
```diff
@@ -171,11 +171,11 @@ bool HDMapManager::set_map_router_lanelet(const lanelet::LaneletMapPtr & lanelet
...
-bool HDMapManager::get_origin_from_filename(std::string filename, lanelet::Origin & origin)
+bool HDMapManager::get_origin_from_filename(std::string filename, lanelet::Origin * origin)
 {
   if (originList.find(filename) == originList.end()) {
     return false;
   }
-  origin = originList[filename];
+  *origin = originList[filename];
   return true;
 }
```
Edwardius: the use of a pointer here is very odd... you should be able to pass by reference

src/world_modeling/hd_map/include/hd_map/hd_map_manager.hpp
```diff
@@ -96,7 +96,7 @@ class HDMapManager
    * @return whether the projection was successful
    */
   bool project_osm_to_lanelet(
-    std::string filename, const lanelet::Projector & projector, const lanelet::LaneletMapPtr & lanelet_ptr);
+    std::string filename, const lanelet::Projector & projector, lanelet::LaneletMapPtr * lanelet_ptr);
```
Edwardius: generally, passing by pointer is not normal. They are typically reserved for optional parameters or ownership transfer

src/world_modeling/hd_map/include/hd_map/hd_map_manager.hpp
```diff
@@ -85,7 +85,7 @@ class HDMapManager
    * @param lanelet_ptr the LaneletMapPtr where the result is stored
    * @return whether the projection was successful
    */
-  bool project_osm_to_lanelet(std::string filename, lanelet::Origin origin, const lanelet::LaneletMapPtr & lanelet_ptr);
+  bool project_osm_to_lanelet(std::string filename, lanelet::Origin origin, lanelet::LaneletMapPtr * lanelet_ptr);
```
Edwardius: pass by pointer

src/world_modeling/hd_map/include/hd_map/hd_map_manager.hpp
```diff
@@ -113,7 +113,7 @@ class HDMapManager
    * @param origin the origin for the map
    * @return whether the origin was successfully retrieved
    */
-  bool get_origin_from_filename(std::string filename, const lanelet::Origin & origin);
+  bool get_origin_from_filename(std::string filename, lanelet::Origin * origin);
```
Edwardius: pass by pointer

src/world_modeling/localization/include/localization/odom.hpp
```diff
@@ -31,7 +31,8 @@
 
 #include "carla_msgs/msg/carla_ego_vehicle_status.hpp"
 
-using std::chrono_literals;
+using std::chrono_literals::operator""s;
+using std::chrono_literals::operator""ms;
```
Edwardius: best to avoid using

src/world_modeling/world_modeling_bringup/launch/world_modeling.launch.py
```diff
@@ -0,0 +1,46 @@
...
+                PathJoinSubstitution(
+                    [
+                        FindPackageShare("localization"),
+                        "launch",
+                        "localization.launch.py",
+                    ]
+                )
+            ]
+        )
+    )
```
Edwardius: combine the launch files into one, you shouldnt be pointing to two different launch files here, unnessesary

Edwardius (changes requested):
generally, the linting issue you see is not fixed by turning whatever into pointers. You have to go into the code and understand if you are modifying the arguments you are passing in or not. If the funciton is, its ok to pass as a non-const reference (ie. pcl::PointCloud<PointT> & ground) . If the function is not, pass it as a const reference.

docker/world_modeling/world_modeling.Dockerfile
Edwardius: @brian386 @lucasreljic here because its relevant to perception ^^

src/world_modeling/hd_map/src/hd_map_manager.cpp
jstephenhuang: Thats the thing I wanted to tell you, the cpplinting doesnt like non-const reference variables, and since in this case yes it will be absolutely better to use pass by reference, cpp linting will complain as there is no const after the reference variable to allow it to be modified. Using pass by pointer is the only way I found to avoid linting issues and compiling issues. Maybe we should change the linting a bit to allow non const reference variable?

docker/world_modeling/world_modeling.Dockerfile
lucasreljic: Currently we have model weights in wato-drive2/perception_weight/ and we just mount the model to the container. So yea a script that automatically pulls from Wato-drive2 to an assets directory based on what you're running could be useful. That would make it easier for incoming members cause Wato-drive2 is not very organized

src/world_modeling/world_modeling_bringup/launch/world_modeling.launch.py
jstephenhuang: I combined most of the launch files in the bringup launch files so the launch file in the other nodes can be deleted, but I left it there. I want to be able to cleanly merge the other working branches for wm before deleting them... After I will do huge clean up...

docker/world_modeling/world_modeling.Dockerfile
Edwardius: yeah, i think a separate curl is okay, from either wato-drive or google drive (which is more accessible to people who just dev on their own pc)

PR #219, opened 2025-09-05
Quick Fixes for CARLA ROS Bridge

VishGit1234:
### 📑  Description

- CARLA ROS Bridge Dockerfile is broken, mainly due to ROS Foxy reaching EOL (end-of-life)
  - This PR modifies the dockerfile (will admit it is a bit hacky) to get the image building again
- Fixes small bug in one of the custom nodes we wrote for the ROS bridge 

### ✅  Checklist


docker/simulation/carla_ros_bridge/carla_ros_bridge.Dockerfile
```diff
@@ -40,23 +41,23 @@ RUN apt-get update -qq && \
...
-        ros-humble-derived-object-msgs \
-        ros-humble-cv-bridge \
-        ros-humble-vision-opencv \
-        ros-humble-rqt-image-view \
-        ros-humble-rqt-gui-py \
+        ros-foxy-tf2-geometry-msgs \
+        ros-foxy-tf2-eigen \
+        ros-foxy-ackermann-msgs \
+        ros-foxy-derived-object-msgs \
+        ros-foxy-cv-bridge \
```
Edwardius: if foxy is reaching EOL, why we downgrading to foxy?

VishGit1234: We're not downgrading, it was always foxy there, but I think when doing the pre-commit change the specific lines you've highlighted were mistakenly changed from foxy to humble.

Edwardius: ah ic, are there no newer versions if carla ros bridge?

Edwardius: looking at https://github.com/carla-simulator/ros-bridge I dont see why we couldnt upgrade the rosbridge, as well as download all the dependencies with rosdep instead of direct apt installation

VishGit1234: i've tried it with humble and everything kind of breaks, and in the docs they specify ros2 foxy as the version.

Edwardius (approved):

PR #229, opened 2025-10-03
deleted dockerfiles/nodes and added perception launch

brianzheng206:
### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456


src/perception/perception_bringup/launch/perception.launch.py
lucasreljic, line 1: Looks good, since camera_object detection is gonna be in deep_ros, this will have to be changed, but for now this is ok.

modules/docker-compose.perception.yaml
lucasreljic, line 1: Probably should add a mount to wato-drive2 for the models, which can be a general mount to the perception_models folder.

lucasreljic (approved):
Looks good, with this revamp we will eventually have to add deep_ros into perception bringup and find a way to call it from one launch file. For now though we can keep it like this

modules/docker-compose.perception.yaml
brianzheng206: done 😄

Edwardius (approved):
LGTM!

PR #232, opened 2025-10-09
Add Skeleton Folder for Prediction (and Kalman Filter node)

yashj19:
### 📑  Description

Added a new prediction module in src and created a skeleton Kalman filter node within it.

### ✅  Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ wato_monorepo_v2
 │   └── world_modeling
 └── watod
 ```
+
 ### Interfacing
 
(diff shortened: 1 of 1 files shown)


docker/prediction/prediction.Dockerfile
Edwardius, line 1: too much! prediction should just be part of world modeling

Edwardius: you should launch the prediction node under world modeling bringup

PR #234, opened 2025-10-09
Add eve_description package

thomasonzhou:
### 📑  Description
Add URDF and yaml launch file for nominal extrinsics

### ✅  Checklist

### 📝  Notes for reviewers
TODO: Fix/remap sensor frames to be consistent with URDF


src/interfacing/eve_description/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
Edwardius: this shouldn't be needed

PR #251, opened 2025-11-20
Hot Fix Bag recorder

Edwardius:
### 📑  Description

Bag recorder was stopping prematurely. To counteract this, a trap was added to explicitly handle SIGTERM and SIGINT to safely close the bag recorder (limiting the chances of file corruption). Also the network of the bad recorder had to be the same as the network of the other containers.

### ✅  Checklist


thomasonzhou (approved):
+1

PR #280, opened 2026-01-14
Irene hdmap

irenexu1:
### 📑  Description
 Removal of regulatory element processing functions, added documentation for the hd map node, and created services for closest lanelet to either a GPS point, XYZ point, or bounding box. Also a lane semantics and objects in lane, where it gives info on the number of adjacent lanes, speed limits etc and the existence of bike lanes, cross walks within a certain distance.
### ✅  Checklist

### 📝  Notes for reviewers
Unsure if ringroad.osm is entirely correct. The node returns a lot of warnings for this map for unidentified objects (likely referencing various regulatory elements or objects in the map such as trees). The default map is set as ringroad currently so if a lot of warnings pop up during the foxglove bridge launch, this might be why

### DIFF
--- a/src/world_modeling/hd_map/include/hd_map/hd_map_router.hpp
+++ b/src/world_modeling/hd_map/include/hd_map/hd_map_router.hpp
@@ -84,7 +84,7 @@ public:
   /**
   * Convert local XYZ coordinates to GPS coordinates.
   *
-  * @param point3d the filename of the map
+  * @param point3d XYZ coordinates
   * @return GPS coordinates
   */
   lanelet::GPSPoint project_point3d_to_gps(lanelet::BasicPoint3d point3d);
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+src/world_modeling/hd_map/src/hd_maps/ringroad.osm
```
Edwardius, line 1: oddly specific for gitignore

src/world_modeling/hd_map/config/params.yaml
```diff
@@ -3,7 +3,7 @@
 hd_map_service:
   ros__parameters:
     version: 1
-    osm_map_filename: "boston-seaport.osm"
+    osm_map_filename: "ringroad.osm"
```
Edwardius, line 6: where is this file stored?

PR #282, opened 2026-01-17
Setup Simulation for Planning

Edwardius:
### 📑  Description

Sets up simulation to be usable with our planning stack for testing purposes.

https://github.com/user-attachments/assets/2314cf2a-427b-4040-ba03-fd2b25bc34b7

### ✅  Checklist

### DIFF
--- a/watod-config.sh
+++ b/watod-config.sh
@@ -36,7 +36,7 @@
 ##   - "interfacing:dev"                  : Interfacing in dev mode (editable)
 ##   - "interfacing:dev perception:dev"   : Both in dev mode
 
-export ACTIVE_MODULES="simulation"
+# export ACTIVE_MODULES=""
 
 ############################## ADVANCED CONFIGURATIONS ##############################
 ## Name to append to docker containers. DEFAULT = "<your_watcloud_username>"


modules/docker-compose.yaml
```diff
@@ -37,40 +37,13 @@ services:
...
-      dockerfile: docker/template.Dockerfile
-      args:
-        MODULE_SOURCE: "${INFRASTRUCTURE_IMAGE:?}:source_${TAG}"
-        MODULE_DEPS: "${INFRASTRUCTURE_IMAGE:?}:deps_${TAG}"
-      cache_from:
-        - "${INFRASTRUCTURE_IMAGE:?}:${TAG}"
-        - "${INFRASTRUCTURE_IMAGE:?}:main"
-      target: deploy
-    image: "${INFRASTRUCTURE_IMAGE:?}:${TAG}"
-    command: bash -c "ros2 run topic_tools relay /tf /tf_new & ros2 launch foxglove_bridge foxglove_bridge_launch.xml port:=${FOXGLOVE_BRIDGE_PORT:?} send_buffer_limit_bytes:=${FOXGLOVE_BRIDGE_SEND_BUFFER_LIMIT_BYTES:?} && fg"
```
VishGit1234, line 63: Is this foxglove buffering not a problem anymore?

modules/docker-compose.yaml
```diff
@@ -272,87 +245,81 @@ services:
...
         condition: service_started
         required: false
 
 #################### SIMULATION SERVICES ####################
   carla_sim:
-    image: carlasim/carla:0.9.13
+    image: carlasim/carla:0.10.0
+    runtime: nvidia
+    init: true
+    network_mode: host
```
VishGit1234: Won't setting network mode as host cause port conflict when two different people try and up the simulation stack on the server?

docker/template.Dockerfile
```diff
@@ -82,6 +82,7 @@ RUN apt-get update && \
     curl \
     wget \
     htop \
+    nano \
```
VishGit1234, line 85: do we need nano here?

modules/docker-compose.yaml
```diff
@@ -272,87 +245,81 @@ services:
...
-        - "${SIMULATION_CARLA_ROS_BRIDGE_IMAGE:?}:${TAG}"
-        - "${SIMULATION_CARLA_ROS_BRIDGE_IMAGE:?}:main"
+        - "${SIMULATION_IMAGE:?}:${TAG}"
+        - "${SIMULATION_IMAGE:?}:main"
       target: deploy
-    image: "${SIMULATION_CARLA_ROS_BRIDGE_IMAGE}:${TAG}"
+    image: "${SIMULATION_IMAGE}:${TAG}"
     environment:
       - CARLA_HOSTNAME=${COMPOSE_PROJECT_NAME:?}-carla_sim-1
       - USE_ACKERMANN_CONTROL=False
```
VishGit1234: i think this env var can be deleted right?

src/simulation/carla_ros_bridge/carla_bringup/config/carla_bridge.yaml
```diff
@@ -0,0 +1,145 @@
+# CARLA ROS2 Bridge Configuration
+
+# Scenario Server Configuration
+# See https://carla.readthedocs.io/en/latest/adv_synchrony_timestep/ for timing settings
+scenario_server:
+  ros__parameters:
+    carla_host: "localhost"
+    carla_port: 2000
+    carla_timeout: 10.0
```
VishGit1234: Instead of repeating these 3 params for every node, couldn't you just put them at the top with wildcard (`/**`) to have them apply to all nodes?

modules/docker-compose.yaml
VishGit1234, line 257: Won't setting network mode as host cause port conflicts for people running on the server? I understand for ROS it's not a problem since you set the ROS_DOMAIN_ID, but that doesnt apply for the sim container.

VishGit1234 (changes requested):
Mainly glanced through the code but everything seems to look great!

I added a few comments, but really the only main one is about potential port conflicts for the sim container.

modules/docker-compose.yaml
VishGit1234: I think you just need to set the sim port in `watod-setup-env.sh` similar to what's being done for the foxglove bridge

modules/docker-compose.yaml
Edwardius: launched within the launchfile now

docker/template.Dockerfile
Edwardius: since its a dev environment, we can install any useful dev tools like nano

modules/docker-compose.yaml
Edwardius: theres some client side port management that also needed to be managed, but yeah addressed

PR #304, opened 2026-01-25
Behaviour Tree V1

JStephenHuang:
### 📑  Description

A MVP of the behaviour tree to tackled two main behaviour:
- lane navigation
- intersections (traffic lights, stop sign, yield)

(refer to the xml file for the design of the tree, i can send excalidraw if needed)

https://github.com/user-attachments/assets/27bf6194-57cd-47e0-a6b4-043a58a4092e

https://github.com/user-attachments/assets/c93ea018-05c1-4e2e-9564-3f38b06c10b8

### ✅  Checklist

### 📝  Notes for reviewers
This is in no way a ROBUST behaviour tree. A lot has not been implemented (e.g. fallbacks for failed nodes, fallbacks for unintended behaviours). 

Furthermore, the logic for some of the nodes are a little scuffed and could be more robust. Notably:
- get intersection context
- active traffic control element passed

I won't give the full shpill, you can try looking at the logic and you can see that it is definitely not the best way to achieve the intended behaviour. This was my + chat's best idea/implementation constrained by what world modeling had to offer. You will see that a lot of the behaviour tree logic assumes a lot of stuff (i.e that the car will always be on the route), if not then behaviour tree kinda crashes out. See **scuffed behaviours**.

I think the biggest problem with my implementation is that I kinda limit the car's context to the Lanelet found under the global route. In other words, I get the ego's context based on where it is on the global route (e.g. getting the traffic reg elem that ego is currently dealing with is by looking through the lanelets ahead + the current lanelet in the global route).

**note**: I did try my best to explain the logic of the heavy nodes.

TL;DR: The design, the implementation definitely needs a second opinion.

However, for now, I think this is good enough to at least try to integrate with lattice planner, if you see in the video (and trying your best to decipher logs), the bt is decent a spitting the correct lane behaviour a some parts (when the car is on the route).

### Scuffed behaviours 

- SetRoute setting the global route starting from the oncoming lane
- Behavhiour tree might still be latched on traffic reg elem that is nowhere close to the car.

### DIFF
--- a/config/foxglove_config/all_cams_compressed.json
+++ b/config/foxglove_config/all_cams_compressed.json
@@ -1,0 +1,693 @@
+{
+  "configById": {
+    "Image!3apg1a1": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_lower_ne/image_rect_compressed",
+        "calibrationTopic": "/camera_lower_ne/camera_info"
+      }
+    },
+    "Image!2441es7": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_lower_nw/image_rect_compressed",
+        "calibrationTopic": "/camera_lower_nw/camera_info"
+      }
+    },
+    "Image!266npvt": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_lower_se/image_rect_compressed",
+        "calibrationTopic": "/camera_lower_se/camera_info"
+      }
+    },
+    "Image!1i31mjw": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_lower_sw/image_rect_compressed",
+        "calibrationTopic": "/camera_lower_sw/camera_info"
+      }
+    },
+    "Image!nd40y3": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_ee/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_ee/camera_info"
+      }
+    },
+    "Image!1w6bw23": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_ne/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_ne/camera_info"
+      }
+    },
+    "Image!10griz9": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_nn/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_nn/camera_info"
+      }
+    },
+    "Image!11g52fu": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_nw/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_nw/camera_info"
+      }
+    },
+    "Image!19gf33s": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_se/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_se/camera_info"
+      }
+    },
+    "Image!x2bmzq": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_ss/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_ss/camera_info"
+      }
+    },
+    "Image!2ajz6yx": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_sw/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_sw/camera_info"
+      }
+    },
+    "Image!1805klt": {
+      "cameraState": {
+        "distance": 20,
+        "perspective": true,
+        "phi": 60,
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOffset": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "thetaOffset": 45,
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {},
+      "layers": {},
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {
+        "imageTopic": "/camera_pano_ww/image_rect_compressed",
+        "calibrationTopic": "/camera_pano_ww/camera_info"
+      }
+    },
+    "3D!3iebfs2": {
+      "cameraState": {
+        "perspective": true,
+        "distance": 211.71135530822306,
+        "phi": 52.63502454992829,
+        "thetaOffset": 122.77414075286605,
+        "targetOffset": [
+          11.937569913836406,
+          2.0797627684185094,
+          -1.0351921873990263e-19
+        ],
+        "target": [
+          0,
+          0,
+          0
+        ],
+        "targetOrientation": [
+          0,
+          0,
+          0,
+          1
+        ],
+        "fovy": 45,
+        "near": 0.5,
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {
+        "/lidar_cc/velodyne_points": {
+          "visible": true,
+          "colorField": "ring",
+          "colorMode": "colormap",
+          "colorMap": "turbo"
+        },
+        "/lidar_ne/velodyne_points": {
+          "visible": true,
+          "colorField": "intensity",
+          "colorMode": "flat",
+          "colorMap": "turbo",
+          "flatColor": "#eddb80ff"
+        },
+        "/lidar_nw/velodyne_points": {
+          "visible": true,
+          "colorField": "intensity",
+          "colorMode": "flat",
+          "colorMap": "turbo",
+          "flatColor": "#ed8080ff"
+        }
+      },
+      "layers": {
+        "grid": {
+          "visible": true,
+          "drawBehind": false,
+          "frameLocked": true,
+          "label": "Grid",
+          "instanceId": "b2241ef7-06c8-4b6f-861e-62e3472986f1",
+          "layerId": "foxglove.Grid",
+          "size": 10,
+          "divisions": 10,
+          "lineWidth": 1,
+          "color": "#248eff",
+          "position": [
+            0,
+            0,
+            0
+          ],
+          "rotation": [
+            0,
+            0,
+            0
+          ]
+        }
+      },
+      "publish": {
+        "type": "point",
+        "poseTopic": "/move_base_simple/goal",
+        "pointTopic": "/clicked_point",
+        "poseEstimateTopic": "/initialpose",
+        "poseEstimateXDeviation": 0.5,
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      },
+      "imageMode": {},
+      "followTf": "base_footprint"
+    }
+  },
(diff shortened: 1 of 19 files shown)


docker/world_modeling.Dockerfile
```diff
@@ -8,7 +8,13 @@ FROM ${BASE_IMAGE} AS source
 
 WORKDIR ${AMENT_WS}/src
 
-# Copy in source code
+# Behaviour Tree ROS2
+RUN git clone https://github.com/BehaviorTree/BehaviorTree.ROS2.git --branch humble behaviortree_ros2
+WORKDIR ${AMENT_WS}/src/behaviortree_ros2
+RUN git checkout 6c6aa078ee7bc52fec98984bed4964556abf5beb
+WORKDIR ${AMENT_WS}/src
```
Edwardius: behaviourtree.cpp can become a vendor package, you can see how thats done in patchworkspp pkg in the monorepo

src/interfacing/INTERFACING_README.md
Edwardius, line 1: looks like your commit history is messed up. generally rebase is very hard to manage, and can result in unwanted changes from past commits. To stop this, i usually use `git pull origin main` without the rebase option. If this stays, then you might need to create a new PR and branch that is most up to date with main.

src/world_modeling/behaviour/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+  behaviortree_cpp
+  behaviortree_ros2
+  world_model_msgs
+  behaviour_msgs
+  geometry_msgs
+  nav_msgs
+  lanelet_msgs
+  tf2_ros
+  tf2_geometry_msgs
+)
```
Edwardius: ament_target_dependencies is deprecated, you should ur target_link_libraries instead (try to make things PRIVATE not PUBLIC) PRIVATE means the library is included in your header file

src/world_modeling/behaviour/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+  DESTINATION share/${PROJECT_NAME}
+)
+
+# ══════════════════════════════════════════════════════════════════════
+# Testing
+# ══════════════════════════════════════════════════════════════════════
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
```
Edwardius: we dont use ament_lint

src/world_modeling/behaviour/include/behaviour/behaviour_tree.hpp
```diff
@@ -0,0 +1,92 @@
...
+  {
+    return blackboard_->get<T>(key);
+  }
+
+private:
+  void registerNodes();
+  void buildTree(const std::string & tree_file_path);
+
+  rclcpp_lifecycle::LifecycleNode::SharedPtr lifecycle_node_;
+  rclcpp::Node::SharedPtr ros_node_;
```
Edwardius: aha, yeah this might be an issue, i think its actually better if we just not make it lifecycle managed. Which actually makes sense to me because BT should actually be what *manages* all the lifecycle managers, who then manage the nodes. As a result, BT is the all knowing and can manage everything from route managements to fault fallback and node lifecycles.

src/world_modeling/behaviour/behaviour/include/behaviour/dynamic_object_store.hpp
Edwardius, line 1: we can discuss what extra services you need from WM

src/world_modeling/behaviour/include/behaviour/interfaces/timers/tick_tree_timer.hpp
Edwardius, line 1: actually, the way i implemented BT in the past was that this tick timer was actually just part of BT node, not a separate thing. Im not too sure how this is being instantiated atm though, so if these two ideas are equivalent then ignore me

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/actions/determine_lane_behaviour.hpp
Edwardius, line 1: should be documented better, docstrings are good

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/actions/determine_reg_elem_node.hpp
```diff
@@ -0,0 +1,82 @@
...
+namespace behaviour
+{
+/**
+   * @class DetermineRegElemAction
+   * @brief BT node to detect upcoming regulatory elements (intersections, traffic lights, etc.)
+   * based on the current lane context and upcoming path.
+   *
+   * TODO(wato): This node needs significant improvement. Deducing reg elem types and IDs
+   * from lanelet flags is fragile. We need better world_model services that can provide
+   * a structured "Regulatory Context" for the current route.
```
Edwardius: make sure you are making note of these todos

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/actions/execute_behaviour_action.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "behaviour_msgs/action/execute_behaviour.hpp"
+
+namespace behaviour
+{
+
+/**
+   * @class ExecuteBehaviourAction
+   * @brief BT node to command the vehicle to execute a specific behavior (lane follow, lane change, etc.)
+   *
+   * TODO(wato): action server on lattice planner
```
Edwardius: atm, lattice planner is assuming its getting a topic of the state it should be in. If its not possible to repetitively send out a state topic, then what can do instead is have a service topic that changes the state the lattice should be in (ie. lane change service request -> lattice updates its internal scoring system -> bt sees successful state change and conitnues)

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/actions/set_wall_service.hpp
Edwardius, line 1: oh FUCK i forgot i need to add this, will do

src/world_modeling/behaviour/behaviour/include/behaviour/utils/custom_types.hpp
```diff
@@ -0,0 +1,139 @@
...
+using RegulatoryElementArray = std::vector<RegulatoryElement>;
+using RegulatoryElementArrayPtr = std::shared_ptr<RegulatoryElementArray>;
+
+/**
+   * @brief Lane behaviour enum for specifying driving maneuvers.
+   *
+   * Can be used in XML like: behaviour="follow_lane"
+   */
+
+// TODO(wato): Figure out how to register enums with btcpp
```
Edwardius: what is this issue? do the enums not exist in the msg? so you can just import those

src/world_modeling/behaviour/launch/behaviour.launch.py
Edwardius, line 1: use yaml launchfiles

src/world_modeling/behaviour/src/behaviour_node.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+  auto ego_pose_period = std::chrono::milliseconds(static_cast<int64_t>(1000.0 / ego_pose_rate_hz));
+  auto tick_period = std::chrono::milliseconds(static_cast<int64_t>(1000.0 / rate_hz));
+
+  // subbers
+  interfaces_.push_back(std::make_unique<interfaces::LaneContextSubscriber>(this, tree_));
+  interfaces_.push_back(std::make_unique<interfaces::GoalPointSubscriber>(this, tree_));
+  interfaces_.push_back(std::make_unique<interfaces::DynamicObjectsSubscriber>(this, dynamic_object_store_));
+
+  // TODO(wato): Add OdometrySubscriber here once implemented the publisher
```
Edwardius: you need this? cant you just use TF

src/world_modeling/behaviour/behaviour/src/behaviour_tree.cpp
Edwardius, line 1: this is really the only ugly bit, all those factory registrations. However, this is a fundamental design flaw of behaviourtree.CPP, something which people with more OCD than me want to fix https://github.com/BehaviorTree/BehaviorTree.CPP/issues/1040

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/actions/execute_behaviour_action.hpp
jstephenhuang: the execute behaviour node will now be publishing a topic to /execute_behaviour and it will have custom msg type ExecuteBehaviour.msg (just a string for now, but we can add more info in the future)

src/world_modeling/behaviour/behaviortree_ros2_vendor/package.xml
```diff
@@ -0,0 +1,18 @@
...
+<package format="3">
+  <name>behaviortree_ros2_vendor</name>
+  <version>0.0.0</version>
+  <description>ROS2 vendor package for the BehaviorTree.ROS2 library.</description>
+  <maintainer email="js3huang@watonomous.ca">jstephhuang</maintainer>
+  <license>Apache-2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>behaviortree_cpp</depend>
```
Edwardius, line 12: heh? is this an existing rosdep package for behaviortree? in that case theres no point to this vendor package

src/world_modeling/behaviour/behaviour/config/params.yaml
Edwardius, line 1: Im surprised theres so little params here, i was expecting some more for timeouts, minor tweaks to behaviours, etc. or is that determined in the xml? if thats the case, then i think we need to define a clear separation of concerns: what params are tuned in the BT xml file and what params are tuned in the rosnode. Personally, id maybe populate the blackboard with a ros param server so that all the other BT nodes can initialize themselves accordingly............................... i think. Let me know what you think

src/world_modeling/behaviour/behaviour/include/behaviour/dynamic_object_store.hpp
```diff
@@ -0,0 +1,217 @@
...
+// Thread-safe cache of dynamic objects with fast lookup indexes.
+//
+// Logic:
+// - Build an immutable `Snapshot` from each incoming world-object message.
+// - Classify each object into supported entity types (car, traffic light).
+// - Populate lookup indexes by object id and by lanelet id.
+// - Atomically replace the current snapshot under a mutex.
+//
+// Assumptions:
+// - `detection.results[0]` is the best class prediction when results exist.
```
Edwardius: why is a cache nessesary? my understanding is that BT functions in the present, not the past.

is this because dynamic objects isnt a service? would it be better if there was a service to get dynamic objects so that you can call to get them during an action that needs them to reason on?

I just believe that it feels like an architectural oversight that we have two threadsafe stores for dynamic objects, one inside world_model and one here

src/world_modeling/behaviour/behaviour/include/behaviour/area_occupancy_store.hpp
Edwardius, line 1: similar thing here, as my comments on dynamic occupancy store.

BT should just take a snapshot of this from world_model. Otherwise, its a waste of resources.

src/world_modeling/behaviour/behaviour/include/behaviour/behaviour_node.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+     * @brief Timer callback to look up the latest TF transform and updates the blackboard.
+     */
+  void tfTimerCallback();
+
+  // Core component
+  std::shared_ptr<BehaviourTree> tree_;
+  std::shared_ptr<DynamicObjectStore> dynamic_objects_store_;
+  std::shared_ptr<AreaOccupancyStore> area_occupancy_store_;
```
Edwardius: see comment about these, this feels like a architectural blunder

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+     * @brief Ticks the behavior tree.
+     */
+void BehaviourNode::tickTreeTimerCallback()
+{
+  try {
+    // send latest world state snapshots to blackboard
+    tree_->updateBlackboard("dynamic_objects.snapshot", dynamic_objects_store_->snapshot());
+    tree_->updateBlackboard("area_occupancy.snapshot", area_occupancy_store_->snapshot());
```
Edwardius: i see... ok services for these is more important for you, otherwise we have a duplication of concerns

Edwardius: for reference,  these can be replaced with something like `snapshot = dynamic_objects_service_client ->send_request()`

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
```diff
@@ -0,0 +1,203 @@
...
+/**
+     * @brief Fetches latest TF transform and updates the BT blackboard with ego state (velocity and position).
+     */
+void BehaviourNode::tfTimerCallback()
+{
+  try {
+    const auto tf = tf_buffer_->lookupTransform(map_frame_, base_frame_, tf2::TimePointZero);
+    const rclcpp::Time current_time = tf.header.stamp;
+
+    // 1. Extract Position as geometry_msgs::msg::Point
```
Edwardius: 1. , 2. , ... is an obvious sign of LLM generated code :P

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
```diff
@@ -0,0 +1,203 @@
...
+    ego_point->y = tf.transform.translation.y;
+    ego_point->z = tf.transform.translation.z;
+
+    // Convert to TF2 types for math
+    tf2::Vector3 current_position(ego_point->x, ego_point->y, ego_point->z);
+    tf2::Quaternion current_orientation;
+    tf2::fromMsg(tf.transform.rotation, current_orientation);
+
+    // 2. Calculate Velocity (Twist)
+    geometry_msgs::msg::Twist::SharedPtr ego_velocity = std::make_shared<geometry_msgs::msg::Twist>();
```
Edwardius: whats this for? should get this from odometry msg

src/world_modeling/behaviour/behaviour/src/behaviour_tree.cpp
```diff
@@ -0,0 +1,77 @@
...
+  intersection_registrar.register_nodes(factory_, params);
+  lane_navigation_registrar.register_nodes(factory_, params);
+}
+
+void BehaviourTree::buildTree(const std::string & tree_file_path)
+{
+  blackboard_ = BT::Blackboard::create();
+
+  // Set the node on the blackboard so custom nodes can access it
+  blackboard_->set<rclcpp::Node::SharedPtr>("node", node_);
```
Edwardius, line 66: is there a way to avoid this? ie, the other nodes initialize with the ros node handle in their constructor?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/execute_behaviour_publisher.hpp
Edwardius, line 1: am i tripping, are you using ExecuteBehaviour.action? or just ExecuteBehaviour.msg

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/get_lanelets_by_reg_elem_service.hpp
Edwardius, line 1: whats wrong with having the dynamic objects and area occupancy be similar to this?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
```diff
@@ -0,0 +1,134 @@
...
+    if (!tf_buffer) {
+      RCLCPP_ERROR(logger(), "[%s] missing input port: tf_buffer", name().c_str());
+      setOutput("error_message", "missing_port:tf_buffer");
+      return false;
+    }
+
+    const std::string base_frame = ports::tryGet<std::string>(*this, "base_frame").value_or("base_link");
+
+    const double width = ports::tryGet<double>(*this, "width").value_or(5.0);
+    const double length = ports::tryGet<double>(*this, "length").value_or(1.0);
```
Edwardius: there a reason why these defaults are set here? should be paramaterized?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    auto tf_buffer = ports::tryGetPtr<tf2_ros::Buffer>(*this, "tf_buffer");
+    if (!tf_buffer) {
+      RCLCPP_ERROR(logger(), "[%s] missing input port: tf_buffer", name().c_str());
+      setOutput("error_message", "missing_port:tf_buffer");
+      return false;
+    }
+
+    const std::string base_frame = ports::tryGet<std::string>(*this, "base_frame").value_or("base_link");
```
Edwardius: problem with base link is that the wall is constantly the same distance away from the car. if the car moves, the wall moves with it

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/ego_stopped_condition.hpp
```diff
@@ -0,0 +1,72 @@
...
+    return {
+      BT::InputPort<double>("ego_velocity"),
+      BT::InputPort<double>("threshold_velocity"),
+    };
+  }
+
+  BT::NodeStatus tick() override
+  {
+    auto ego_vel = ports::tryGet<double>(*this, "ego_velocity");
+    const double thresh = ports::tryGet<double>(*this, "threshold_velocity").value_or(0.1);
```
Edwardius: hidden magic number?

Edwardius: are these values set in the xml?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/ego_stopped_condition.hpp
Edwardius, line 1: i think this goes back to what I said about ego_velocity already. Odometry should be used, not a custom interpolation logic

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/goal_reached_condition.hpp
Edwardius, line 1: the car currently doesn't have logic to exist a lanelet and further optimize itself to some goal point, i think this should track the goal lanelet reached instead

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/actions/get_stop_line_pose_action.hpp
Edwardius, line 1: is this for wall of doom?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/actions/get_intersection_context_action.hpp
Edwardius, line 1: question, for a given intersection, is BT reconstructing the whole intersection? doesn't it actually only care about a subset of the lanelets in that intersection?

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/conditions/is_traffic_light_state_condition.hpp
```diff
@@ -0,0 +1,80 @@
...
+      std::cerr << "[IsTrafficLightState] Missing expected." << std::endl;
+      return BT::NodeStatus::FAILURE;
+    }
+
+    // case-insensitive comparison
+    auto norm = [](std::string s) {
+      std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+      return s;
+    };
+
```
Edwardius: unexpected util out of nowhere, lol

Edwardius: what is this trying to solve?

src/world_modeling/world_modeling_bringup/launch/world_modeling.launch.yaml
```diff
@@ -208,6 +208,20 @@ launch:
...
+  #     exec: behaviour_node
+  #     name: behaviour_node
+  #     namespace: ""
+  #     output: screen
+  #     param:
+  #       - from: $(find-pkg-share behaviour)/config/params.yaml
+  #     remap:
+  #       - from: execute_behaviour
+  #         to: /behaviour/execute_behaviour
+
```
Edwardius, line 256: launch it!

src/world_modeling/behaviour/behaviour/README.md
Edwardius, line 1: we're gonna need more information than this lol, also add a DEVELOPING.md. you can reference the doc structure of simulation/carla_ros_bridge

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/actions/get_stop_line_pose_action.hpp
jstephenhuang: yes

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
parthpkale: I believe the values you are referring to (5.0 or 1.0) are fallback values incase tryGet is unable to fetch the parameterized values.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/ego_stopped_condition.hpp
parthpkale: Yes, ideally odometry should be subscribed to directly for getting stopped_condition - could/should be changed.

src/world_modeling/behaviour/behaviortree_ros2_vendor/package.xml
jstephenhuang: yes behaviourtree_cpp is a rosdep package, but the wrappers for the behaviourtree_ros2, that is only accessible through the repo: https://github.com/BehaviorTree/BehaviorTree.ROS2

and that repo depends on behaviourtree_cpp.

src/world_modeling/behaviour/behaviour/config/params.yaml
jstephenhuang: yea honeslty did not do my due diligence when it came to params, was just eager to test with lattice planner. However I will add more params, there are definitely some nodes that can use it.

In terms of your idea of populating the behaviour tree with a ros param server, if you look at the registar.hpp under the common folder, you can see that I do tweak their timeout duration, for each individiual server node. But yes, the milliseconds are hard coded and I will have them as params in the config.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
jstephenhuang: will paramaterized and be would be tuneable in the config file.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
jstephenhuang: can the spawn wall use pose from from the map frame? since when i saw your video, I saw that you spawned the wall using base_link (i might be wrong). 

But I'm a little confused given a point (in the map) and I transform that point to the base frame and I spawn the wall, you are saying that the wall will move with it? Even if I spawn the wall just once?

Will using map coordinates not make the wall move?

src/world_modeling/behaviour/behaviour/config/params.yaml
jstephenhuang: https://github.com/WATonomous/wato_monorepo/blob/eaac8b289d0e7e9d2376c519613e4cb1e1897ea8/src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/registrar.hpp#L49-L67

src/world_modeling/behaviour/behaviour/include/behaviour/dynamic_object_store.hpp
jstephenhuang: Yes it was because dynamic objects was a topic, it is possible to use the btros2 wrapper BtRosTopicSub, but that would still make an internal store and acting in the past. Since I would be saving the objects, but it wouldn't listen to new objects until all children succeed. 

The only way to get messages up to date, is to run that node in parallel or reactively (re runs children on tick). The slight issue I had with this was that I would have to search through the raw WorldObjectArray.msg everytime I wanted to check for dynamic objects. 

I wanted to standardize (and only implement the code once) how I retrieved useful dynamic objects based on useful queries (id or lanelet_id). And I think it works out? because on every tick, I take a snapshot of all dynamic objects that I was able to listen and I have the methods to retrieve the objects I want.

If i was s service with the useful queries, for me it would make more sense (for me) to convert it into a node. But honestly I can try making a BtRosTopicSub and seeing how that works. Let me know.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/conditions/is_traffic_light_state_condition.hpp
jstephenhuang: yea i dont think this is useful, this was just make sure when we compare traffic light states that we don't have case sensitive comparison. but yea i just refer the the traffic light state from your world model (lower case snake_case)

src/world_modeling/world_modeling_bringup/launch/world_modeling.launch.yaml
jstephenhuang: okok!

src/world_modeling/behaviour/behaviour/README.md
jstephenhuang: holy, yea I forgot about that part, will make a juicy dev md.

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
jstephenhuang: yea if that was the case it would be a littler better, but also it would be fire to have the useful queries/triage done for me as well. Since finding all objects on some lanelet requires some work, but also for the type and mainly deciphering the detection 3d.

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
jstephenhuang: :p

will clean up

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
jstephenhuang: I really just need the velocity of the ego, but I swear last time I was listening to a topic for the velocity using a (placeholder topic /velocity_odom) and you said I could get that from the tf? I might be misunderstanding. But yea I really just need to velocity of the car to check if its stopped.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/execute_behaviour_publisher.hpp
jstephenhuang: just ExecuteBehaviour.msg. I left ExecuteBehaviour.action if ever we wanted to go back to ros actions. But i can def remove.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/get_lanelets_by_reg_elem_service.hpp
jstephenhuang: ok let me see if i can implement a btrostopicsub and see how it plays out, it would require some changed in the design tho.

src/world_modeling/behaviour/behaviour/src/behaviour_tree.cpp
jstephenhuang: this is actual not necessary for the ros node wrappers, since they are passed in the RosNodeParam, but this for any node that would want to use ros node methods (logger or some other method)

but currently no bt node uses this value in the blackboard, so it can be removed. I was just thinking that maybe in the future some general bt node (not ros wrapper) would need it.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/ego_stopped_condition.hpp
jstephenhuang: yes it is set in the xml, but will have a param in the config for the default value

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/goal_reached_condition.hpp
jstephenhuang: hmmm, currently the bt just tells the car to safely follow the global route that it sets given the goal point. I agree that the bt is kinda rudimentary right now, and we can add some behaviour to overtake. But I would want to see the bt actual work with lattice planner before adding more.

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/intersection/actions/get_intersection_context_action.hpp
jstephenhuang: no it is not reconstructing the whole intersection. This node's job is to find the reg elem that we are approaching so that we can appropriately handle it. The way it does it, is by looking at the global route and checking a some fixed lanelets ahead (using distance)

src/world_modeling/behaviour/behaviortree_ros2_vendor/package.xml
Edwardius: ok, sounds good

src/world_modeling/behaviour/behaviour/config/params.yaml
Edwardius: oh ic, yeah these should be moved

src/world_modeling/behaviour/behaviour/src/behaviour_node.cpp
Edwardius: yeah you can listen to /ego/odom for odometry which gives you a twist you canuse

Edwardius: for sim only, this needs to be a remap in the launchfile so that we can swap between the topic we listen in sim vs real life

src/world_modeling/behaviour/behaviour/src/behaviour_tree.cpp
Edwardius: you can also pass around just the logger through constructors as well. but yeah sg

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/execute_behaviour_publisher.hpp
Edwardius: rm

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/get_lanelets_by_reg_elem_service.hpp
Edwardius: on my end, i can expose two services for you from world_model, that way you get exactly what you need at a specific time

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
Edwardius: ok sg, yeah my concern is that we actually arent passing anything to these values, and they are just defaulting all the time (secret magic number)

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/ego_stopped_condition.hpp
Edwardius: /ego/odom (should be a remap)

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/conditions/goal_reached_condition.hpp
Edwardius: i mean near the end of a route, where it needs to pull over or something

src/world_modeling/behaviour/behaviour/include/behaviour/nodes/common/actions/spawn_wall_service.hpp
Edwardius: you'll have to spawn the wall in map frame, so that it stays in place when the car moves, the demo was just showing you that you can choose any frame of ref (though, it was probably misleading, sorry). You'll prolly need to do a couple transforms of your own (ie. you want a wall to spawn 2m in front of the car, get that pose in base_link frame and then apply a transform, and then send it to our service to spawn a wall of doom)

src/world_modeling/behaviour/behaviour/include/behaviour/dynamic_object_store.hpp
Edwardius: i think a service call should at least remove the extra cache logic here, and all you'll have to do is parse it accordingly to what you need

Edwardius: ill have to create those services, on my radar today

Edwardius (approved):
LGTM! I think its better that we add in whats already written here instead of keeping a 10 000 line pr open. @parthpkale you can pull in main and then i can relook at your PR

PR #305, opened 2026-01-25
[BUG] Dockerfile messing with permissions in devcontainer for watcloud

JStephenHuang:
### 📑  Description

When building and running devcontainer on watcloud, you would not have permissions to edit any of the content that was mounted on to the devcontainer. But you had sudo rights, so you can pass the ownership to the user (bolty) that was created for the devcontainer, but that then lead the user BACK ON THE HOST not have any permissions to edit any of the files that was mounted by the devcontainer. Essentially, your files are not yours anymore and are locked because some user owns the code from a devcontainer. 

You will see weird stuff like this:

```bah
rm: cannot remove 'wato_monorepo_old2/src/world_modeling/occupancy/launch/occupancy.launch.py': Permission denied
```

Supposed to be: 

```
drwxr-xr-x  13 jstephhuang jstephhuang   20 Jan 25 00:01 wato_monorepo
```

but ends up being:

```bash
drwxr-xr-x  13    32279796    32279796   20 Jan 25 00:01 wato_monorepo
```

### ✅  Checklist

### 📝  Notes for reviewers

IMPORTANT: this is a workaround!

It works on dev now, you can edit both on the devcontainer and the host without any permissions issue. However, before in the devcontainer the user would be:

```bash
bolty@trpro-slurm2:~/ament_ws$ 
```

now it is just back to what it was before:
```bash
root@trpro-slurm1:/home/bolty/ament_ws#
```

### DIFF
--- a/docker/template.Dockerfile
+++ b/docker/template.Dockerfile
@@ -68,6 +68,7 @@ FROM build AS deploy
 # Source Cleanup, Security Setup, and Workspace Setup
 RUN rm -rf "${AMENT_WS:?}"/* && \
     chown -R "${USER}":"${USER}" "${AMENT_WS}"
+USER ${USER}
 
 ################################ Develop ################################
 FROM rosdep_install AS develop


docker/template.Dockerfile
```diff
@@ -66,8 +66,7 @@ RUN . "/opt/ros/${ROS_DISTRO}/setup.sh" && \
 FROM build AS deploy
 
 # Source Cleanup, Security Setup, and Workspace Setup
-RUN rm -rf "${AMENT_WS:?}"/* && \
-    chown -R "${USER}":"${USER}" "${AMENT_WS}"
+RUN rm -rf "${AMENT_WS:?}"/*
```
Edwardius: this shouldn't affect anything, since its for non-devcontainers

docker/template.Dockerfile
```diff
@@ -86,8 +85,9 @@ RUN apt-get update && \
     tree
 
 # Make ament_ws owned by bolty
-RUN chown -R "${USER}":"${USER}" "${AMENT_WS}"
-USER ${USER}
+# TODO(wato): (dont make ament_ws be owned bolty or it will clash with user permissions on watcloud)
+# RUN chown -R "${USER}":"${USER}" "${AMENT_WS}"
+# USER ${USER}
```
Edwardius: you can completely delete this, its ok

PR #315, opened 2026-01-28
John/prediction vehicles

jonz9:
### 📑  Description
Implemented kinematic bicycle model with pure pursuit and vehicle hypotheses generation in prediction node.

### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456

### 📝  Notes for reviewers
changed the output format for TrajectoryHypothesis so it uses std_msgs::msg::Header and geometry_msgs::msg::PoseStamped. updated for vehicle hypotheses generation but not constant velocity model or pedestrian model.

For initial_state velocity, currently set to detection.results[0].hypothesis.score; for now as a placeholder, until object tracking velocity can be used.

### DIFF
--- a/src/world_modeling/prediction/src/motion_models.cpp
+++ b/src/world_modeling/prediction/src/motion_models.cpp
@@ -96,10 +96,10 @@ std::vector<geometry_msgs::msg::PoseStamped> BicycleModel::generateTrajectory(
   }
 
   KinematicState current_state = initial_state;
-  double t = 0.0;
+  double t = dt;
   const double lookahead_distance = 3.0;
 
-  while (t < horizon) {
+  while (t <= horizon) {
     // compute heading with pure pursuit logic
     double min_distance = std::numeric_limits<double>::max();
     size_t closest_idx = 0;
@@ -213,9 +213,9 @@ std::vector<geometry_msgs::msg::PoseStamped> ConstantVelocityModel::generateTraj
   std::vector<geometry_msgs::msg::PoseStamped> trajectory;
 
   KinematicState current_state = initial_state;
-  double t = 0.0;
+  double t = dt;
 
-  while (t < horizon) {
+  while (t <= horizon) {
     current_state = propagate(current_state, dt);
 
     geometry_msgs::msg::PoseStamped pose_stamped;


src/world_modeling/prediction/src/trajectory_predictor.cpp
```diff
@@ -91,34 +96,146 @@ std::vector<TrajectoryHypothesis> TrajectoryPredictor::generateVehicleHypotheses
...
+  initial_state.y = detection.bbox.center.position.y;
 
-  for (double t = 0.0; t <= prediction_horizon_; t += time_step_) {
-    geometry_msgs::msg::Pose waypoint;
-    waypoint.position.x = current_x + velocity * std::cos(heading) * t;
-    waypoint.position.y = current_y + velocity * std::sin(heading) * t;
-    waypoint.position.z = 0.0;
-    waypoint.orientation.w = 1.0;
+  if (!detection.results.empty() && detection.results[0].hypothesis.score > 0.0) {
+    initial_state.v = detection.results[0].hypothesis.score;
```
RyanL2: Not sure if the velocity should be taking values from the hypothesis.score

src/world_modeling/prediction/src/motion_models.cpp
```diff
@@ -40,36 +45,135 @@ KinematicState BicycleModel::propagate(const KinematicState & initial_state, dou
...
+  std::vector<geometry_msgs::msg::PoseStamped> trajectory;
+
+  // If no path provided, return empty trajectory
+  if (path_points.empty()) {
+    return trajectory;
+  }
 
   KinematicState current_state = initial_state;
   double t = 0.0;
+  const double wheelbase = 2.5;  // Default wheelbase
```
RyanL2: wheelbase is defined in motion_models.hpp already, change this to remove redundant declaration.

src/world_modeling/prediction/src/trajectory_predictor.cpp
RyanL2: The waypoints and timestamps fields have been removed from the struct. Please update this to use the new poses field (std::vector<PoseStamped>) so the build will compile.

RyanL2: Same as line R265.

RyanL2: Also please update trajectory_predictor.hpp:52-55 header comment as it references old format.

src/world_modeling/prediction/src/trajectory_predictor.cpp
jonz9: Right, forgot I left that as a placeholder. Should I just set that as a default velocity value for now? How do we get the object velocity? I don't think it's a field in the Detection3D input array.

src/world_modeling/prediction/src/motion_models.cpp
jonz9: changed wheelbase to a private variable and declared at bicycle model initialization

src/world_modeling/prediction/src/trajectory_predictor.cpp
jonz9: fixed

jonz9: fixed

src/world_modeling/prediction/src/motion_models.cpp
```diff
@@ -39,37 +45,134 @@ KinematicState BicycleModel::propagate(const KinematicState & initial_state, dou
...
+    for (size_t i = 0; i < path_points.size(); ++i) {
+      double dx = path_points[i].x() - current_state.x;
+      double dy = path_points[i].y() - current_state.y;
+      double distance = std::sqrt(dx * dx + dy * dy);  // finding Euclidean distance
+
+      if (distance < min_distance) {
+        min_distance = distance;
+        closest_idx = i;
+      }
+    }
```
Edwardius: whats the point of a pure pursuit controller here?

src/world_modeling/prediction/src/motion_models.cpp
```diff
@@ -39,37 +45,133 @@ KinematicState BicycleModel::propagate(const KinematicState & initial_state, dou
...
+    pose_stamped.pose.position.y = current_state.y;
+    pose_stamped.pose.position.z = 0.0;
 
-    trajectory.push_back(pose);
+    pose_stamped.pose.orientation.w = std::cos(current_state.theta / 2.0);
+    pose_stamped.pose.orientation.x = 0.0;
+    pose_stamped.pose.orientation.y = 0.0;
+    pose_stamped.pose.orientation.z = std::sin(current_state.theta / 2.0);
+
+    trajectory.push_back(pose_stamped);
```
RyanL2, line 174: t starts at 0 but the state is already propagated by dt before stamping, so every pose is timestamped one step behind its actual time. start t at dt to match what the pedestrian/cyclist code does.

PR #316, opened 2026-01-28
feature: oscc interfacing 

FowlFarmer:
### 📑  Description

! Note on the large diff: The OSCC library which is no longer maintained is now added within the oscc interfacing node. It has been stripped of the firmware and hardware folders and only retains the CPP apis necessary for interfacing with ROS. Disregard any diffs in a file specifically within interfacing/oscc_interfacing/oscc/.

ROS2 node that interfaces with the physical OSCC boards through OSCC CAN API

This node does these things:

- Subscribes to /joystick/roscco

- Publishes to /oscc_interfacing/is_armed (Just a bool, 100HZ)

- Publishes to /oscc_interfacing/wheel_speeds (4 floats, one per wheel)
- Publishes to /oscc_interfacing/steering_wheel_angle (float, degrees, 0 = centered)
  - You can model this with ackermann reference frames and get an odom for
  - speed and angular velocity for localization

- Is a server for the service /oscc_interfacing/arm
  - Attempt to either arm or disarm.
  - SetBool service: true = attempt to arm, false = attempt to disarm
  - Returns success/fail (message empty)
  
### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,7 +40,6 @@ repos:
     hooks:
       - id: clang-format
         args: ["--style=file:.config/clang-format"]
-        exclude: "^(src/embedded/)"
 
   # C++ linting
   - repo: https://github.com/cpplint/cpplint
@@ -48,8 +47,7 @@ repos:
     hooks:
       - id: cpplint
         args: ["--config=.cpplint.cfg", "--filter=-runtime/references", --quiet, --output=sed]
-        exclude: "^(src/embedded/)"
-
+        exclude: "^(src/embedded/|src/interfacing/oscc_interfacing/oscc/)"
   # CMake linting
   - repo: https://github.com/cmake-lint/cmake-lint
     rev: 1.4.3


src/interfacing/joystick_interfacing/config/joy_config.yaml
```diff
@@ -6,31 +6,18 @@ joystick_node:
...
-    #ROSCCO SCALLING ---- max speed and steering angle
-    roscco_max_speed: 1.0
-    roscco_max_steering_angle: 0.3
-
+    #max speed and steering angle
+    max_speed: 1.0
+    max_steering_angle: 0.3
     #invert steering and throttle
     invert_steering: true
     invert_throttle: false
```
shawnytron, line 32: I think this here is out of date, try merging from latest main. there should be some merge conflicts to resolve. 

this max speed and Max steering angle got parsed into 2 different scaling variable, one for Ackerman and the other for oscc

src/interfacing/oscc_interfacing/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+  if(wato_test_FOUND)
+    add_wato_test(oscc_interfacing_test test/oscc_interfacing_test.cpp
+      LIBRARIES oscc_interfacing
+    )
+    target_link_libraries(oscc_interfacing_test
+      ${sensor_msgs_TARGETS}
+      ${std_msgs_TARGETS}
+      roscco_msg::roscco_msg__rosidl_typesupport_cpp
+      std_srvs::std_srvs__rosidl_typesupport_cpp
+    )
```
shawnytron: I think this can be removed since you linked it above and it's public

PR #339, opened 2026-02-07
added fixes for steering wheel scaling and systemctl startup config script for can

shawnytron:
This PR is based on the oscc_interfacing PR -- contains fixes for tuning steering tomorrow

should be merged into main once oscc_interfacing is merged 
 
### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456


src/interfacing/pid_control/include/pid_control/pid_control_node.hpp
Edwardius, line 1: i wouldn't make these have a default, it should only read from the config

PR #341, opened 2026-02-08
Local Planner V1

harsharan-r:
### 📑  Description
Added a lattice planner to compute a local path for the car to follow, while considering obstacles and the global route(bt).

https://youtu.be/TOSzz5OhVPU (video of car driving around and generated path in orange)
| Before BT Route Set | After BT Route Set |
|:----------------------:|:-----------------:|
| image1|image2 |

### ✅  Checklist

### 🔗  Related Issues / PRs
Depends on #304 

### 📝  Notes for reviewers
This is an initial implementation of a local planner with some known simplifications made. Planning is done directly in the global frame rather than Frenet coordinates; while Frenet would allow for better cost modelling (especially lateral deviation from the centerline), the conversion overhead was avoided in this iteration. The cost function currently uses the global occupancy grid for obstacle avoidance. This assumes it is a map of static obstacles, and should be removed if no map is implemented. Additionally, the lateral deviation penalty is approximated using path curvature rather than actual distance from the centerline, which works but isn't ideal. 

This is not the most robust planner, as there is significant deviation from the route during some curves; however, it should work for a track environment. The behaviour tree was pulled from #304 but will only be tested in a stationary scenario, as Carla doesn't yet follow the route.


modules/docker-compose.dev.yaml
Edwardius, line 1: prolly remove this,

src/action/action_bringup/launch/action_launch.py
Edwardius, line 1: that mpc node is not useful anymore, also if you can, please turn action_bringup's launchfile into a yaml formatted launchfile

src/action/local_planning/config/local_planner_params.yaml
```diff
@@ -0,0 +1,33 @@
+local_planner_node:
+  ros__parameters:
+    
+    # Topic names
+    lanelet_ahead_topic: "/world_modeling/lanelet/lanelet_ahead"
+    odom_topic: "/ego/odom"
+    costmap_topic: "/world_modeling/costmap"
+    bt_topic: "/behaviour/execute_behaviour"
```
Edwardius: topic's are set through remaps inside the launchfile. This is so that people can quickly orchestrate new launchfile setups's dealing with plumbing in the process.

keep these topics as genmeric as posible. ie "odometry" or "lanelet_ahead"

src/action/local_planning/config/local_planner_params.yaml
```diff
@@ -0,0 +1,33 @@
...
+    
+    # Publishing
+    planned_paths_vis_topic: "planned_paths_markers"
+    final_paths_vis_topic: "final_path_markers"
+    path_topic: "path"
```
Edwardius: ditto above

src/action/lattice_planning/include/lattice_planning/lattice_planning_core.hpp
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <vector>
+
+#include <eigen3/Eigen/Core>
+#include <eigen3/Eigen/Dense>
+
+// TODO(wato) make this a true frenet point and make converter functions
```
Edwardius: ?

src/action/local_planning/include/local_planning/local_planner_core.hpp
```diff
@@ -0,0 +1,62 @@
...
+    int steps
+  );
+
+  // Eigen::Matrix4d compute_jacobian(
+  //   const Eigen::Vector3d& p, 
+  //   const Eigen::Vector4d& error, 
+  //   const PathPoint& start, 
+  //   const PathPoint& target, 
+  //   int steps
+  // );
```
Edwardius: rm

src/action/lattice_planning/include/lattice_planning/lattice_planning_core.hpp
Edwardius, line 1: plz add docstring

src/action/lattice_planning/include/lattice_planning/lattice_planning_node.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include <optional>
+
+#include "local_planning/local_planner_core.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp_lifecycle/lifecycle_node.hpp"
+#include "rclcpp_lifecycle/lifecycle_publisher.hpp"
+
+#include <visualization_msgs/msg/marker_array.hpp>
+#include <visualization_msgs/msg/marker.hpp>
```
Edwardius: i've generally kept foxglove-related marker stuff separate from the node itself. My suggestion is a separate node that reads the path coming out of lattice_planner and publish a marker for foxglove.

src/action/lattice_planning/include/lattice_planning/lattice_planning_node.hpp
Edwardius, line 1: just from a glance, how separate is node and core logic? ie. i wouldn't expect the node class to have something like cost computations, since thats not really ros specific

src/action/local_planning/include/local_planning/local_planner_node.hpp
```diff
@@ -0,0 +1,139 @@
...
+  // subscriber callbacks
+  void lanelet_update_callback(const lanelet_msgs::msg::LaneletAhead::ConstSharedPtr & msg);
+  void update_vehicle_odom(const nav_msgs::msg::Odometry::ConstSharedPtr & msg);
+  void update_costmap(const nav_msgs::msg::OccupancyGrid::ConstSharedPtr & msg);
+  void set_preferred_lanes(const behaviour_msgs::msg::ExecuteBehaviour::ConstSharedPtr & msg);
+  
+  // publisher wrappers
+  void publish_final_path(const Path & path);
+  void publish_planned_paths_vis(const std::vector<Path> & paths);
+  void publish_final_path_vis(const Path & path);
```
Edwardius: define private functions before private members

Edwardius: convention reasons, nothing else

src/action/local_planning/src/local_planner_core.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+std::vector<PathPoint> LocalPlannerCore::generate_path(
+  PathPoint start, 
+  PathPoint target,
+  PathGenParams pg_params)
+{
+  std::vector<PathPoint> path;
+
+  double norm = std::numeric_limits<double>::infinity();
+  double spiral_coeffs[4];
```
Edwardius: these sort of magic numbers that are purely for mathematical computations should be defines as constants

src/action/local_planning/src/local_planner_core.cpp
```diff
@@ -0,0 +1,213 @@
...
+               normalise_angle(final_state.theta - target.theta);
+      
+      norm = error.norm();
+
+      if (norm < pg_params.tolerance) break;
+
+      Eigen::Vector3d params(spiral_params[1], spiral_params[2], spiral_params[4]);
+      
+      // TODO(wato) ideally add the curvature as a parameter as well
+      //    currently having problems with short arc distanes
```
Edwardius: ? this will be forgotten into the depths of time

src/action/local_planning/src/local_planner_core.cpp
```diff
@@ -0,0 +1,213 @@
...
+double LocalPlannerCore::normalise_angle(double angle){
+    return std::remainder(angle, 2.0 * M_PI);
+}
+
+void LocalPlannerCore::calculate_spiral_coeff(const double p[5], double (&coeffs)[4]){
+  coeffs[0] = p[0];
+  coeffs[1] = (-11 * p[0] + 18 * p[1] - 9 * p[2] + 2 * p[3])  / (2 * p[4]);
+  coeffs[2] = (9 * (2 * p[0] - 5 * p[1] + 4 * p[2] - p[3]))   / (2 * p[4] * p[4]);
+  coeffs[3] = (-9 * (p[0] - 3 * p[1] + 3 * p[2] - p[3]))      / (2 * p[4] * p[4] * p[4]);
+}
```
Edwardius: what is this black magic? are these values conventional? or are they the result of a precomputed optimization that occurred somewhere else

src/action/local_planning/src/local_planner_core.cpp
```diff
@@ -0,0 +1,213 @@
...
+  double spiral_params[5];
+
+  // Initial arc length estimate
+  double sf_est = get_euc_dist(start.x, start.y, target.x, target.y);
+  
+
+  // Set intial spiral parameters
+  spiral_params[0] = start.kappa;
+  spiral_params[1] = start.kappa + (target.kappa - start.kappa) * 0.33;  // 1/3 point
+  spiral_params[2] = start.kappa + (target.kappa - start.kappa) * 0.67;  // 2/3 point
```
Edwardius: can't you write 1/3 and 2/3?

src/action/lattice_planning/include/lattice_planning/lattice_planning_core.hpp
Edwardius: especially the path generation loop

src/action/lattice_planning/src/lattice_planning_node.cpp
```diff
@@ -0,0 +1,565 @@
...
+  cm_params.preferred_lane_cost = this->declare_parameter("cm_preferred_lane_cost",20.0);
+  cm_params.unknown_occupancy_cost = this->declare_parameter("cm_unknown_occupancy_cost",50.0);
+  cm_params.max_curvature_change = this->declare_parameter("cm_max_curvature_change",0.1);
+
+  //  - Path Generation -
+  pg_params.max_iterations = this->declare_parameter("max_iterations", 20);
+  pg_params.steps = this->declare_parameter("path_steps", 20);
+  pg_params.tolerance = this->declare_parameter("convergence_tolerance", 0.25);
+  pg_params.newton_damping = this->declare_parameter("newton_damping", 0.7);
+  pg_params.max_step_size = this->declare_parameter("max_step_size", 1.0);
```
Edwardius: parameters go in on_configure

src/action/local_planning/src/local_planner_node.cpp
```diff
@@ -0,0 +1,565 @@
...
+  double cm_h  = costmap->info.height;
+  
+  // Preferred Lane Cost
+  if(!preferred_lane){
+    path_cost += params.preferred_lane_cost;
+  }
+
+  for(const auto & pt: path.path){
+    // TODO(wato) get rid of occupancy cost if no static costmap is implemented
+    // Occupancy Cost
```
Edwardius: also will be forgotten in the depths of time

src/action/local_planning/src/local_planner_node.cpp
```diff
@@ -0,0 +1,565 @@
...
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn LocalPlannerNode::on_shutdown(
+  const rclcpp_lifecycle::State &)
+{
+  RCLCPP_INFO(this->get_logger(), "Shutting down Local Planning node");
+  lanelet_ahead_sub_.reset();
+  costmap_sub_.reset();
+  bt_sub_.reset();
+  RCLCPP_INFO(this->get_logger(), "Node shut down");
+  return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+}
```
Edwardius: i usually have these at the start

src/action/local_planning/src/local_planner_node.cpp
```diff
@@ -0,0 +1,565 @@
...
+int main(int argc, char * argv[])
+{
+    rclcpp::init(argc, argv);
+    auto lc_node = std::make_shared<LocalPlannerNode>(rclcpp::NodeOptions());
+    lc_node->configure();
+    lc_node->activate();
+    rclcpp::spin(lc_node->get_node_base_interface());
+    rclcpp::shutdown();
+    return 0;
+}
```
Edwardius: this is not needed if you are registering it as an rclcpp_component

src/action/local_planning/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+
+add_library(local_planner_core SHARED
+  src/local_planner_core.cpp
+)
+target_include_directories(local_planner_core PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
+)
+# Put only what local_planner_core uses:
+ament_target_dependencies(local_planner_core)
```
Edwardius: ament_target_dependencies is deprecated, use target_link_libraries instead

src/action/local_planning/package.xml
```diff
@@ -3,20 +3,29 @@
 <package format="3">
   <name>local_planning</name>
   <version>0.0.0</version>
-  <description>TODO(wato): Package description</description>
-  <maintainer email="e23zhou@watonomous.ca">eddyzhou</maintainer>
-  <license>Apache 2.0</license>
+  <description>TODO: Package description</description>
+  <maintainer email="todo@todo.todo">todo</maintainer>
+  <license>TODO: License declaration</license>
```
Edwardius: fill out, hello@watonomous.ca maintainer WATonomous or your name

src/action/local_planning/package.xml
```diff
@@ -3,20 +3,29 @@
...
-  <test_depend>ament_copyright</test_depend>
-  <test_depend>ament_flake8</test_depend>
-  <test_depend>ament_pep257</test_depend>
-  <test_depend>python3-pytest</test_depend>
+  <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>visualization_msgs</depend>
-  <depend>geometry_msgs</depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
```
Edwardius: rm, we use pre-commit

.gitignore
```diff
@@ -13,6 +13,8 @@ ssh_config/
 watod-config.local.sh
 watod-config.sh
 
+local_planning_copy
+
```
Edwardius: ?

Edwardius (changes requested):
Nice work!!!

ocd, but i think we should rename this to frenet_lattice_planner or something smart like that lol. local_planner isn't really that accurate.

also, where are docssssssa;kldj;aldj3od. Docstrings, and please limit docs to important things only, no code samples. I do still want to know what lattice planner does briefly through the readme or something.

src/action/lattice_planning/src/lattice_planning_node.cpp
Edwardius: ignore this

src/action/local_planning/src/local_planner_core.cpp
harsharan-r: these are like the derived equations to get the coefficients of a cubic sprial using the curvaturae parameters, so you gotta compute these alot for different curvatures

src/action/local_planning/src/local_planner_core.cpp
harsharan-r: These change multiple times per path because you gotta iteratively get to the cubic sprial so these change multiple during a path generation.

src/action/local_planning/src/local_planner_core.cpp
harsharan-r: Got rid of the comment, the paths seem to be generating find without optimizing curvature through the path

src/action/lattice_planning/include/lattice_planning/lattice_planning_core.hpp
harsharan-r: All functionality seems to be preserved without using frenet so we don't need the change. Nuked the comment

src/action/lattice_planning/include/lattice_planning/lattice_planning_node.hpp
harsharan-r: I can do that for the final path, but not to visualize the paths for each terminal point; the only point of that visualization is for debugging. Should I nuke it all? Also, the path msg can natively be visualized in foxglove so should I add a marker node for it?

src/action/local_planning/src/local_planner_core.cpp
Edwardius: the size of the array changes?

src/action/local_planning/src/local_planner_core.cpp
Edwardius: do we ever have to change these coefficient values?

Edwardius: they feel very arbitrary

src/action/local_planning/src/local_planner_core.cpp
harsharan-r: Ohhhh , yea that doesn't change that makes sense, will change

src/action/local_planning/src/local_planner_core.cpp
harsharan-r: No, those values should never be touched; those are derived from the cubic spiral equation. Here is the source, its on the second page [lattice planner](https://www.ri.cmu.edu/pub_files/2011/5/20100914_icra2011-mcnaughton.pdf)

Edwardius (approved):

PR #344, opened 2026-02-08
added pedestrian prediction

girish316:
### 📑  Description
Implemented multi-hypothesis pedestrian trajectory prediction in the TrajectoryPredictor.
The pedestrian model now generates behavior-aware trajectories (confident walk, hesitant/wandering, and stop) using a constant-velocity baseline with bounded uncertainty.

### ✅  Checklist
- Updated comments where behavior or assumptions are non-obvious
- No existing tests were broken

### 📝  Notes for reviewers
- Replaced the placeholder constant-velocity pedestrian model with behavior-based hypotheses
- Forward motion is accumulated, while lateral deviation is applied relative to the forward path at each timestep. 
- Output format (std_msgs::msg::Header + geometry_msgs::msg::PoseStamped) is unchanged


src/world_modeling/prediction/src/trajectory_predictor.cpp
```diff
@@ -246,44 +246,152 @@ std::vector<TrajectoryHypothesis> TrajectoryPredictor::generateVehicleHypotheses
...
+  // Typical pedestrian walking speed (m/s)
+  const double nominal_walk_speed = 1.4;
+
+  // Precompute orientation aligned with heading (planner-friendly)
+  geometry_msgs::msg::Quaternion aligned_orientation;
+  aligned_orientation.w = std::cos(heading_yaw * 0.5);
+  aligned_orientation.z = std::sin(heading_yaw * 0.5);
+
+  // Hypothesis 1: Confident forward walking
+  TrajectoryHypothesis confident_walk;
```
RyanL2, line 1193: generatePedestrianHypotheses() sets probability (0.55 / 0.30 / 0.15), but processObject() in prediction_node.cpp:159-160 then calls intent_classifier_->assignProbabilities() which overwrites them. 

I was planning for us to not use the intent classifier yet. But since vehicle prediction already used intent classifier, let's refactor the assigning probabilities part to the intent classifier for pedestrians for consistency.

src/world_modeling/prediction/src/trajectory_predictor.cpp
```diff
@@ -246,44 +246,152 @@ std::vector<TrajectoryHypothesis> TrajectoryPredictor::generateVehicleHypotheses
...
+  for (double time_sec = time_step_;
+       time_sec <= prediction_horizon_;
+       time_sec += time_step_) {
+
+    // Slower forward progress due to hesitation
+    accumulated_forward_distance +=
+      0.6 * nominal_walk_speed * time_step_;
+
+    // Lateral uncertainty grows with time (non-cumulative)
+    double lateral_offset = 0.15 * std::sqrt(time_sec);
```
RyanL2: The hesitant walk just drifts to one side. 0.15 * sqrt(t) is always positive and only applied to one side. This looks like it's trying to model uncertainty (sqrt(t) growth) but it's just a single curved path. Add a mirrored right-side hypothesis or rethink how you want to represent the uncertainty here.

RyanL2 (changes requested):
Nice work Girish. Make sure to run pre-commit checks before making the PR. 

Check out https://github.com/WATonomous/wato_monorepo/blob/main/DEVELOPING.md for instructions to run pre-commit.

src/world_modeling/prediction/src/trajectory_predictor.cpp
girish316: Moved probability assignment out of generatePedestrianHypotheses() so that
intent_classifier_->assignProbabilities() handles it, consistent with the
vehicle prediction pipeline.

src/world_modeling/prediction/src/trajectory_predictor.cpp
girish316: Added mirrored hesitant hypotheses (left and right) so the lateral uncertainty is symmetric instead of drifting to one side.

RyanL2 (approved):
good to merge

PR #346, opened 2026-02-09
Attribute Assigner

brianzheng206:
### 📑  Description

- using hsv to extract semantics from 2d object detection
- crops the detection (traffic lights, busses, trucks, cars), uses zero copy, converts to hsv and does color detection
- for cars, it detects 2 blobs of bright pixels and checks what color they are, if its 2 red blobs then its braking and if only one amber/red then its turning.
- Traffic lights are published as 3D bounding boxes using intrinsic/extrinsic and set traffic light size (as 1m tall) for finding 3D position.
- An estimated 3D bounding box for cars (trucks and buses included) is also generated but this is very rudementary as it cannot estimate the orientation of these bounding boxes so this is primarily an additional depth measurement to fuse with spatial association node (2D to 3D). 
- Traffic light color detection defaults to red, it uses pixel count and hue/sat values to determine score of colors.
- All hsv color detection is configurable through the perception_bringup.yaml config file.

https://github.com/user-attachments/assets/2494900b-df06-4e20-a023-0b05adfedc28

3D detection of cars
https://github.com/user-attachments/assets/2314a0b2-f696-4ad4-bf7c-b9f13b94609d

### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456


src/perception/attribute_assigner/config/params.yaml
```diff
@@ -0,0 +1,47 @@
+---
+attribute_assigner_node:
+  ros__parameters:
+    sync_queue_size: 10   # ApproximateTime sync queue size for image + detections
+
+    # Class ID mappings (YOLOv8 COCO: 9=traffic_light, 2=car, 7=truck, 5=bus)
+    traffic_light_class_ids: ["traffic_light", "9"]
+    car_class_ids: ["car", "2", "truck", "7", "bus", "5"]
```
brianzheng206: is it still using the aspect ratios for left signal and right signal or are we not doing that

docker/perception.Dockerfile
```diff
@@ -9,7 +9,7 @@ FROM ${BASE_IMAGE} AS source
 WORKDIR ${AMENT_WS}/src
 
 # Copy in source code needed for perception build
-RUN git clone https://github.com/WATonomous/deep_ros.git deep_ros
+RUN git clone https://github.com/WATonomous/deep_ros.git deep_ros --branch test-fix
```
Edwardius: this needs to be merged?

src/perception/attribute_assigner/config/params.yaml
Edwardius: https://github.com/ros-perception/vision_msgs/blob/ros2/vision_msgs/msg/VisionClass.msg , obj should publish this, and we can get a lookup https://github.com/ros-perception/vision_msgs/blob/ros2/vision_msgs/msg/LabelInfo.msg

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
```diff
@@ -0,0 +1,252 @@
...
+  double car_brake_min_saturation{40.0};
+  /// Car: red hue range for brake lights (same wrapping as traffic lights)
+  double car_red_hue_lo{10.0};
+  double car_red_hue_hi{170.0};
+  /// Car: amber hue range (OpenCV H 0-180) and min saturation/value for turn/hazard
+  double car_amber_hue_lo{12.0};
+  double car_amber_hue_hi{35.0};
+  double car_amber_min_saturation{80.0};
+  double car_amber_min_value{100.0};
+};
```
Edwardius: separate header

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
```diff
@@ -0,0 +1,252 @@
...
+ * Represents the confidence scores for each possible car behavior.
+ * Downstream consumers can inspect these to determine detected behaviors.
+ */
+struct CarAttributes
+{
+  double turning_left{0.0};  ///< Confidence the car is turning left
+  double turning_right{0.0};  ///< Confidence the car is turning right
+  double braking{0.0};  ///< Confidence the car is braking
+  double hazard_lights{0.0};  ///< Confidence the car has hazard lights on
+};
```
Edwardius, line 52: separate header

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
```diff
@@ -0,0 +1,252 @@
...
+   * @return true if the detection is a car or vehicle
+   */
+  bool isCar(const vision_msgs::msg::Detection2D & det) const;
+
+  /**
+   * @brief Get the class_id of the highest-scoring hypothesis.
+   * @param det The detection to inspect
+   * @return The class_id string, or empty string if no hypotheses exist
+   */
+  static std::string getBestClassId(const vision_msgs::msg::Detection2D & det);
```
Edwardius: cant getBestClassId do the work of the other two?

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * @brief Classify car behavior from the cropped image: brake lights and turn/hazard signals.
+   *
+   * Converts crop to HSV once and uses integral images for O(1) region mean calculations.
+   * Bottom region: red hue + high S/V -> braking. Left/right regions: amber hue -> turn or hazard.
+   *
+   * @param crop Cropped image (car/vehicle bbox)
+   * @return Attribute confidences for each car behavior
+   */
+  CarAttributes classifyCarBehavior(const cv::Mat & crop) const;
```
Edwardius: i think these can become a plugin system. if anything a separate object or something. that way iun the future, if we want to add more attributes to assign, we  have a simple and structured way to add more attributes

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
```diff
@@ -0,0 +1,252 @@
...
+   * @param class_id The class identifier string
+   * @param score The confidence score
+   * @return A populated ObjectHypothesisWithPose message
+   */
+  static vision_msgs::msg::ObjectHypothesisWithPose makeHypothesis(const std::string & class_id, double score);
+
+  /// Attribute class_id prefix for traffic light states
+  static constexpr auto kStatePrefix = "state:";
+  /// Attribute class_id prefix for car behaviors
+  static constexpr auto kBehaviorPrefix = "behavior:";
```
Edwardius: can be part of that separated plugin sys

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_node.hpp
```diff
@@ -0,0 +1,294 @@
...
+    const std::unordered_map<std::string, sensor_msgs::msg::CameraInfo> & camera_infos,
+    const builtin_interfaces::msg::Time & stamp) const;
+
+  /**
+   * @brief Convert 3D traffic light detections to a MarkerArray for Foxglove visualization.
+   * @param detections_3d 3D detection array to convert
+   * @return MarkerArray with cube markers colored by traffic light state
+   */
+  visualization_msgs::msg::MarkerArray createTrafficLight3DMarkers(
+    const vision_msgs::msg::Detection3DArray & detections_3d) const;
```
Edwardius: feels likea bleed, i think we can contain traffic light stuff better than just slapping it into the main node class

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_node.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  // Simple caching - stores latest messages
+  std::mutex sync_mutex_;
+  MultiImageMsg::ConstSharedPtr cached_multi_image_;
+  MultiCameraInfoMsg::ConstSharedPtr cached_multi_camera_info_;
+
+  // TF2 for camera transforms
+  std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
+  std::shared_ptr<tf2_ros::TransformListener> tf_listener_;
+
```
Edwardius: // ROS Constructs

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_node.hpp
```diff
@@ -0,0 +1,294 @@
...
+  rclcpp_lifecycle::LifecyclePublisher<deep_msgs::msg::MultiDetection2DArray>::SharedPtr detections_pub_;
+  rclcpp_lifecycle::LifecyclePublisher<visualization_msgs::msg::ImageMarker>::SharedPtr image_markers_pub_;
+  rclcpp_lifecycle::LifecyclePublisher<vision_msgs::msg::Detection3DArray>::SharedPtr traffic_lights_3d_pub_;
+  rclcpp_lifecycle::LifecyclePublisher<visualization_msgs::msg::MarkerArray>::SharedPtr traffic_lights_3d_markers_pub_;
+
+  std::unique_ptr<AttributeAssignerCore> core_;
+
+  rclcpp::QoS subscriber_qos_;  ///< QoS profile for detection subscriber (configured in on_configure)
```
Edwardius: i dont get why claude does `///<`

src/perception/attribute_assigner/src/attribute_assigner_node.cpp
```diff
@@ -0,0 +1,1116 @@
...
+
+    const std::string publisher_reliability = this->get_parameter("qos_publisher_reliability").as_string();
+    const std::string publisher_durability = this->get_parameter("qos_publisher_durability").as_string();
+    const int publisher_depth = this->get_parameter("qos_publisher_depth").as_int();
+    publisher_qos_ = createPublisherQoS(publisher_reliability, publisher_durability, publisher_depth);
+
+    this->declare_parameter<int>("sync_queue_size", 10);
+    this->declare_parameter<double>("sync_max_time_diff_ms", 200.0);
+
+    sync_queue_size_ = this->get_parameter("sync_queue_size").as_int();
```
brianzheng206, line 790: is this being used to sync the timestamps?

src/perception/attribute_assigner/src/attribute_assigner_node.cpp
```diff
@@ -0,0 +1,1116 @@
...
+{
+  multi_image_msg_count_++;
+
+  if (msg->images.empty()) {
+    return;  // Skip empty messages
+  }
+
+  // Simply cache the latest multi-image message
+  std::lock_guard<std::mutex> lock(sync_mutex_);
+  cached_multi_image_ = msg;
```
brianzheng206: is it okay to just use the latest multi-image? do we not need to sync the timestamps

docker/perception.Dockerfile
brianzheng206: is this in main or do we need another pr?

src/perception/attribute_assigner/config/params.yaml
lucasreljic: No, so now it detects blobs of red or amber colors wherever they are in the bounding box, and if there is two blobs then says  its braking while only 1 blob means its turning. It gives a score based on size of blob and pixel count

src/perception/attribute_assigner/src/attribute_assigner_node.cpp
lucasreljic: No, so there is a bug in deep_object_detection that only occurs when you run a rosbag. This bug is that the detections have the timestamps of the msg and the MultiImage itself has a timestamp when the multiImage was created. So essentially a sync wouldn't work, but we could change around deep_object_det to publish the current time instead of the timestamp of the image so something we could do.

docker/perception.Dockerfile
lucasreljic: We need another PR unfortunately

brianzheng206: sounds good, we need to add the camera height and width in the camera info message as well, so lets get this deep ros pr merged first

src/perception/attribute_assigner/DEVELOPING.md
```diff
@@ -0,0 +1,169 @@
...
+   - Per-class dimensions for car, truck, and bus
+2. **Projection clamping**: 3D box dimensions are clamped so their projection does not exceed the 2D bbox
+3. **TF transform**: 3D point is transformed from camera frame to `target_frame` (default: `base_link`)
+4. **Orientation**: Vehicle boxes are oriented to face the ego vehicle (yaw from origin to detection position)
+
+### Accuracy: Traffic Lights vs Vehicles
+
+Traffic light 3D positions are significantly more accurate than vehicle positions because traffic lights are a **standardized physical size** — their real-world dimensions are known and consistent. This means the monocular depth estimate (`depth = real_height * fy / bbox_height_pixels`) is reliable.
+
+Vehicles, on the other hand, vary widely in size. A compact car and a full-size SUV both map to class `car` but differ by over a meter in every dimension. The configurable per-class sizes (car/truck/bus) are averages, so depth estimates will be over- or under-estimated for vehicles that deviate from the assumed dimensions.
```
Edwardius: sounds like slop? i dont see why this is relevant

src/perception/attribute_assigner/DEVELOPING.md
```diff
@@ -0,0 +1,169 @@
...
+|-----------|---------|-------------|
+| `min_detection_confidence` | 0.3 | Minimum YOLO score to process a detection |
+| `traffic_light_strip_margin` | 0.25 | Fraction to ignore on each side of TL crop |
+| `car_real_width/height/length` | 1.8/1.8/6.0 | Car dimensions for depth estimation (m) |
+| `truck_real_width/height/length` | 2.5/3.5/8.0 | Truck dimensions (m) |
+| `bus_real_width/height/length` | 2.5/3.2/12.0 | Bus dimensions (m) |
+| `enable_image_markers` | true | Toggle 2D overlay markers |
+| `enable_3d_markers` | true | Toggle 3D visualization markers |
+| `sync_max_time_diff_ms` | 2000.0 | Max time difference for ApproximateTime sync (ms) |
+| `target_frame` | "base_link" | TF frame for 3D detections |
```
Edwardius: should be higher up in ther readme

docker/perception.Dockerfile
lucasreljic: it is merged in deep_ros now

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_core.hpp
lucasreljic: Switched to a plugin system, where generic functions are in the core, and separate classifier files can be created that specify how to detect that specific object.

src/perception/attribute_assigner/include/attribute_assigner/attribute_assigner_node.hpp
lucasreljic: True, I moved most of the math into core, although still need tf lookups in the node but is much cleaner now. Plus with the plugin system specific objects have most 3D detection logic in their own files.

src/perception/attribute_assigner/DEVELOPING.md
lucasreljic: removed

PR #426, opened 2026-04-03
Changes from Friday Testing

Edwardius:
### 📑  Description

This PR is a conglomeration of changes for filming. There are still a handful of issues with the code that we should address. Of which I will make Issues for for the near future.

### ✅  Checklist


src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -404,7 +427,7 @@ multi_camera_sync:
     sync_max_time_diff_ms: 400.0  # Detections usually lag images by about 50-150 ms
 
     # --- Visualization outputs ---
-    enable_image_markers: false  # 2D image marker overlays
+    enable_image_markers: true  # 2D image marker overlays
```
lucasreljic: This is just for visualization so it should normally be set to false

src/perception/tracking/tracking/src/tracking.cpp
```diff
@@ -282,24 +332,21 @@ void TrackingNode::detectionsCallback(vision_msgs::msg::Detection3DArray::Shared
...
+  } catch (const tf2::TransformException &) {
+    // Fall back to latest transform if the stamped lookup fails
+    try {
+      tf_stamped = tf_buffer_->lookupTransform(output_frame_, msg->header.frame_id, tf2::TimePointZero);
+    } catch (const tf2::TransformException & ex) {
+      RCLCPP_WARN(this->get_logger(), "Transform unavailable: %s. Clearing track state.", ex.what());
+      track_filters_.clear();
+      track_centroids_.clear();
+      return;
+    }
```
lucasreljic, line 361: What is this state? Should it just error out if the stamped transform lookup fails?

lucasreljic (approved):
I mean it was tested on the car so everything should be gtg. Just a few things that were kinda patched up for testing (the EKF prediction in the tracking node) we might have to fix.

src/perception/tracking/tracking/src/tracking.cpp
Edwardius: this one no clue

Edwardius: lol

PR #428, opened 2026-04-06
Spatial association classes

lucasreljic:
### 📑  Description

Added class-aware clustering for spatial_association so that smaller classes (like "person" and "bicycle") have different params for clustering (essentially they are reclustered from the initial clustering pass as that is done before the transform into the camera frame.

Additionally, I noticed that tracking was not picking up the "person" class even though the bounding box was persisting. This was due to the IOU distance metric which only tracked objects for ByteTrack if the previous bounding box overlapped the new one, this is rare for small vertical bounding boxes like "person" though. So I added a class-aware change to tracking that makes a few other ByteTrack instances for these outlier conditions to change from IOU to CIOU (the other classes go through the previous general one). 

https://github.com/user-attachments/assets/636eecaf-3307-431a-addb-a92ac56c7549

### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456

### DIFF
--- a/src/perception/perception_bringup/config/perception_bringup.yaml
+++ b/src/perception/perception_bringup/config/perception_bringup.yaml
@@ -168,7 +168,7 @@ multi_camera_sync:
 /**/tracking_node:
   ros__parameters:
     # --- Temporal tracking behavior ---
-    frame_rate: 9
+    frame_rate: 12
     track_buffer: 120
     track_thresh: 0.1
     high_thresh: 0.15
@@ -178,7 +178,7 @@ multi_camera_sync:
     dist_metric: "IOU"
     # Use lidar_cc when running perception without localization (no map→lidar_cc TF).
     # Set to "map" when running with localization so tracks are in the global frame.
-    output_frame: "lidar_cc"
+    output_frame: "map"
     publish_visualization: true
 
     # --- Per-class tracking overrides (parallel arrays). Empty = single global tracker. ---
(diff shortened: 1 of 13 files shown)


src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -167,7 +168,7 @@ multi_camera_sync:
 /**/tracking_node:
   ros__parameters:
     # --- Temporal tracking behavior ---
-    frame_rate: 12
+    frame_rate: 9
```
Edwardius: whats this change for?

src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -177,9 +178,18 @@ multi_camera_sync:
     dist_metric: "IOU"
     # Use lidar_cc when running perception without localization (no map→lidar_cc TF).
     # Set to "map" when running with localization so tracks are in the global frame.
-    output_frame: "map"
+    output_frame: "lidar_cc"
```
Edwardius: will this be an issue?

Edwardius (approved):
some small comments

src/perception/perception_bringup/config/perception_bringup.yaml
lucasreljic: Yes thank you for catching that gotta change it back to map

Issue #432, opened 2026-05-18
new ringroad rosbags


brianzheng206:
- record ringroad rosbags with eidos pose and aggregated point clouds

lucasreljic:
Ring road rosback with eidos here: https://drive.google.com/drive/u/1/folders/1293dKJLuDN4_5KQgk3ToORRmp8fOuODu

PR #443, opened 2026-05-20
Topics for rosbag

Matthew0803:
### 📑  Description
eidos + lidar agg rosbag topics & visualization steps for old rosbags

### ✅  Checklist

### 🔗  Related Issues / PRs
related to #432


src/world_modeling/eidos/DEVELOPING.md
brianzheng206, line 1: don't think we will need this, we should be recording the ego pose from eidos everytime we record a rosbag in the future. this is just a good to know

src/interfacing/lidar_aggregator/README.md
```diff
@@ -1,2 +1,12 @@
 # Lidar Aggregator
 Package to merge multiple lidar pointclouds together into a single pointcloud.
+
+## using Lidar Aggregator on ROS bags
```
brianzheng206, line 4: same as the other comment, this is good to know. we should be recording this topic everytime we go out for data

brianzheng206 (changes requested):

src/interfacing/lidar_aggregator/README.md
Matthew0803: They're for old rosbags, let me add a note

brianzheng206: im saying we don't need it since we probably wont be using the old rosbags if we get new ones with all the topics we need

brianzheng206 (approved):

PR #455, opened 2026-05-30
fix perception CI

Matthew0803:
### 📑  Description
Perception CI is broken for every PR

### ✅  Checklist

### 🔗  Related Issues / PRs
Broken CI

### DIFF
--- a/.github/templates/check_src_changes/action.yml
+++ b/.github/templates/check_src_changes/action.yml
@@ -19,6 +19,12 @@ runs:
       with:
         files: '!src/**'
 
+    - name: Find changed files inside src/wato_test folder
+      id: changed-files-wato-test
+      uses: tj-actions/changed-files@v42
+      with:
+        files: src/wato_test/**
+
     - name: Find changed files inside infrastructure folders
       id: changed-files-infrastructure
       uses: tj-actions/changed-files@v42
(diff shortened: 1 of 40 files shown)


Matthew0803:
### Problem 1: Every PR triggered perception CI, which always failed

The tracking test was running but crashing partway through because of a bug in the shared test helper (TestExecutorFixture). The bug: when a test finished, it tried to clean up its robot nodes, but the nodes were already gone (destroyed too early).

Fix: The test helper now holds onto every node until cleanup is fully done, then cleans them up in the right order.

### Problem 2: Changing wato_test triggered ALL modules, exposing hidden failures

When we fixed problem 1, the CI detected that wato_test changed and ran tests for every module: action, interfacing, perception, simulation, world_modeling. But those other modules had their own pre-existing failures that nobody had ever seen because nothing ever triggered them before.

Fix: We told the CI to detect changes in wato_test and flag it as a global change (so it correctly runs all modules instead of running nothing at all).

### Problem 3: Action and interfacing CI hung for 60 seconds then got killed

The lifecycle manager tests were getting stuck during cleanup. The test would finish, but the background thread running the robot couldn't figure out it was supposed to stop, so it just sat there. Eventually CI's timer killed it.

Fix: When a test finishes, we now explicitly tell ROS to shut down first. Every waiting operation immediately stops, and the background thread can exit cleanly.

### Problem 4: Perception CI failed because tracking test had no map

The tracking node needed a "map" frame to process detections (it uses this to know where things are in the world). The test never provided one, so every time the node received a detection, it said "I don't know where the map is" and threw it away. The test then timed out waiting for results that never came.

Fix: We added a fake map frame to the test — a simple anchor point saying "map is right here" so the tracking node can process detections normally.

### Problem 5: Simulation CI crashed instantly with no output

The wato_test package (our shared test helper) was being built and tested inside the simulation environment too. When its tests ran, they tried to start ROS twice in the same process, which caused an immediate crash before anything could even print.

Fix: We added a guard that checks "is ROS already running?" before trying to start it again.

.github/templates/check_src_changes/action.yml
```diff
@@ -60,7 +66,7 @@ runs:
     - name: Create list of changed modules
       id: output-changes
       env:
-        ALL_CHANGED: ${{ steps.changed-files-outside-src.outputs.any_changed == 'true' }}
+        ALL_CHANGED: ${{ steps.changed-files-outside-src.outputs.any_changed == 'true' || steps.changed-files-wato-test.outputs.any_changed == 'true' }}
```
lucasreljic: Why is wato-test added here? Should it be added seperately not just in the all_changed as an or statement

lucasreljic (changes requested):
Just the one thing about the github action rest looks good to me

lucasreljic (approved):

PR #458, opened 2026-06-10
Isaac ros camera optimizations

lucasreljic:
Added Isaac ROS to the Perception Container. This means we now avoid copies and keep everything on the GPU, from the camera driver through rectification to compression.*

There is one caveat: no JPEG compression node for Isaac ROS. So currently there is an argument to enable the regular image proc compression node (after Isaac ROS rectification), this allows for disabling it when we no longer need to view JPEG images directly and can feed in the nitros topics directly to downstream object detection (like BEVFusion).

This is the toggle within the `all_cameras_composed.yaml` file
```
  - arg:
      name: "enable_jpeg"
      default: "true"
```

### 📑  Description
This was necessary as it cut down on CPU compute quite a lot (from 70% down to 20%), which means we can now run 8 cameras at 20fps without problems!

This PR also uncommented 5 cameras, so now 8 run on startup.

https://github.com/user-attachments/assets/103bc590-caf6-482e-9fb5-14b977069aef

### ✅  Checklist

### 🔗  Related Issues / PRs
Depends on https://github.com/WATonomous/camera_aravis2_nitros/pull/1

### DIFF
--- a/docker/perception.Dockerfile
+++ b/docker/perception.Dockerfile
@@ -10,7 +10,7 @@ WORKDIR ${AMENT_WS}/src
 
 # Copy in source code needed for perception build
 RUN git clone https://github.com/WATonomous/deep_ros.git deep_ros && \
-    git clone https://github.com/WATonomous/camera_aravis2_nitros.git camera_aravis2_nitros -b nitros-support
+    git clone https://github.com/WATonomous/camera_aravis2_nitros.git camera_aravis2_nitros
 
 COPY src/perception perception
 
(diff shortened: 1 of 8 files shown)


brianzheng206:
Might be out of scope of this PR, do you know how the NITROS pointer works under ipc? ipc passes pointers around, would it just pass a nitros tensor instead? wondering if any testing has been done with the camera nodes, and if it is able to handle the conversion. I could also just be understanding this wrong. otherwise lgtm

brianzheng206 (approved):

lucasreljic:
@brianzheng206 Nitros creates a "GXF" graph under the hood where it manages its own memory pool and negotiates with Nitros nodes on codec types (RGB or NV12). It passes pointers between these nodes and creates blocks so only 1 node changes the memory at a time. Basically, Nitros uses its own form of IPC that is more robust. There are only a handful of msg types that Nitros supports I think RGB8, NV12, and Tensors (for DNN). Isaac ROS has nitros nodes that we can just drop into the launch file to resize, rectify, convert into different types.

PR #466, opened 2026-06-19
oscc graceful disengage

brianzheng206:
### 📑  Description

When autonomy disengages, oscc_interfacing called oscc_disable() immediately on every disengage path. The OSCC steering board stops applying torque the instant it is disabled, so the commanded steering torque dropped from whatever it was (often large, plus the deadzone offset) to zero in a single CAN cycle. The result was a hard step in steering torque that jerked the wheel, and the actuator did not hold any torque during the transition, control snapped back to the operator with no managed handover.

This is unsafe for an operator-in-the-loop handover: the wheel can kick, and on a crowned or curved road the wheel can dart toward center the moment authority is cut.

A graceful disengage was added, only applying to the planned/manual disarm path, where steering torque ramps down over 300ms. Made a simple state machine to track the autonomy states. This is currently local, we should have somewhere publish this state globally

I only applied this behavior in operator override, as we move towards autonomous disengages + gradual handoffs this needs to be applied there as well
### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456

### 📝  Notes for reviewers

Here what needs to be tested before we merge this:
- Disarm mid-corner with held torque → wheel returns smoothly over the ramp window, not in one snap.
- Override / fault / API error during a ramp → boards release instantly (ramp aborts). 
-  roscco commands during DISENGAGING are ignored (throttled warn log).
- enable_graceful_disarm: false and disarm_ramp_ms <= 0 both fall back to instant disarm.


src/interfacing/oscc_interfacing/include/oscc_interfacing/oscc_interfacing.hpp
```diff
@@ -158,9 +202,9 @@ class OsccInterfacingNode : public rclcpp::Node
   rclcpp::TimerBase::SharedPtr event_timer_;
   rclcpp::TimerBase::SharedPtr feedback_timer_;
 
-  // Arm state — protected by arm_mutex_
-  std::mutex arm_mutex_;
```
brianzheng206, line 162: deleted this mutex, none of the roscco callbacks took the lock, i hope this doesnt break anything

FowlFarmer (approved):

PR #467, opened 2026-06-19
seperated topics for joystick idle

### LINKED ISSUE #457: error when switching between control mode
When switching between Ackermann and roscco control mode while the deadman switch is pressed, the steering command will somehow be processed after the deadman switch is released.

Solution could be disabling switching control mode while the deadman switch is pressed

Matthew0803:
### 📑  Description
the root cause of the issue is: the 2 muxes share the same topic for idling. When deadman is pressed (idle = false) and switch to roscco, idle will always be set to true, so no joystick cmd could be processed. The fix is to have 2 topics with same logic to handle ackermann and roscco, while disabling mux switch when deadman is pressed for safety.

### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #457

### DIFF
--- a/src/interfacing/ackermann_mux/DEVELOPING.md
+++ b/src/interfacing/ackermann_mux/DEVELOPING.md
@@ -46,7 +46,7 @@ ackermann_mux:
         topic: /joystick/ackermann
         priority: 200
         has_mask: true
-        mask_topic: /joystick/is_idle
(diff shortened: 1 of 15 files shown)


lucasreljic (approved):
Tested on the car and fixes the switching bug. LGTM

PR #474, opened 2026-06-30
BEVFusion

### LINKED ISSUE #437: cuda-BEVFusion ROS node development
implement the cuda-bevfusion model into the monorepo
- look into how to run the bevfusion model

AshishA26:
### 📑  Description

Implements the bevfusion ROS 2 lifecycle composable node — fuses 6 camera feeds + LiDAR point cloud through a GPU-accelerated TensorRT BEVFusion pipeline to produce 3D bounding boxes.

What's included:

- BEVFusionCore — ROS-free C++/CUDA wrapper around cuda_bevfusion_vendor. Handles TRT engine deserialization (with auto-build from ONNX on first run), FP16 conversion, and inference via a dedicated CUDA stream
- BEVFusionNode — ROS 2 lifecycle node with ApproximateTime sync for camera + LiDAR, camera info caching, TF-based calibration matrix computation, and diagnostic_updater integration
- Publishes vision_msgs/Detection3DArray (transformed into target_frame via TF2) and visualization_msgs/MarkerArray with per-class nuScenes colors
- Optional ring field support for LiDAR (has_ring param), configurable QoS, throttled logging
- Developer docs in bevfusion/DEVELOPING.md and bevfusion/bevfusion/DEVELOPING.md

It has been tested on the car!

https://drive.google.com/file/d/1BAR_-28SGA_m84O2Faobb_fdmMJRZvqH/view?usp=sharing
Note: This is a video from the car testing. At the time this video was made, the orientation and class colors were not fixed. However the PR has these fixes in place now.

### ✅  Checklist

### 📝  Notes for reviewers

- TRT engines are compiled on first run if .plan files are missing from build_dir — first startup takes several minutes
- Model weights live under /opt/watonomous/models/bevfusion/ (mounted by docker-compose, not in this repo). They can be found here https://drive.google.com/drive/folders/1TqwYlcURiw4RxAL2tnRIrzHj7oW7VXze?usp=sharing
- input_multi_image_topic and input_lidar_topic must be changed via params.yaml, not via launch-time remap (message_filters limitation)


AshishA26:
Note that some cameras have been uncommented/commented. Also note that attribute assigner and deep object detection have been commented out for now, but can be uncommented, bevfusion works regardless.

docker/perception.Dockerfile
```diff
@@ -57,3 +57,11 @@ RUN apt-get update && \
       ros-jazzy-isaac-ros-image-proc \
       ros-jazzy-isaac-ros-h264-encoder && \
     rm -rf /var/lib/apt/lists/*
+
+# Install protobuf and git-lfs for cuda_bevfusion_vendor
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    protobuf-compiler \
+    libprotobuf-dev \
+    git-lfs && \
```
lucasreljic, line 66: Can this be added to the above apt-get install? You can add a comment next to it stating the reason why it is used

src/perception/bevfusion/bevfusion/config/params.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+    # Detection confidence threshold — boxes below this score are discarded
+    confidence_threshold: 0.1
+
+    # Camera input resolution (must match the camera hardware)
+    image_width: 1280
+    image_height: 1024
+
+    # Resize limit
+    resize_lim: 0.55
```
lucasreljic, line 36: Im a little confused on this param, is this the resize multiplier or the limit of the minimum resize it can be?

src/perception/bevfusion/bevfusion/config/params.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+    # Model input resolution after resize
+    norm_output_width: 704
+    norm_output_height: 256
```
lucasreljic, line 40: If the above is the multiplier then why is the resolution stated here? I think its best just to put the model height/width instead of having a resize limit/ multiplier

src/perception/bevfusion/bevfusion/config/params.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+    # LiDAR detection zone in meters [x, y, z]
+    min_range: [-54.0, -54.0, -5.0]
+    max_range: [54.0, 54.0, 3.0]
```
lucasreljic, line 44: Why 5 meters into the ground? or is this 5 meters below the lidar frame?

src/perception/bevfusion/cuda_bevfusion_vendor/CMakeLists.txt
```diff
@@ -0,0 +1,229 @@
...
+option(FORCE_BUILD_VENDOR_PKG
+  "Force building cuda-bevfusion from source"
+  OFF)
+
+find_package(ament_cmake REQUIRED)
+
+macro(build_cuda_bevfusion)
+    message(STATUS "Building cuda-bevfusion from source")
+    include(ExternalProject)
+    set(BEVFUSION_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/cuda_bevfusion_install)
```
lucasreljic, line 29: Does this pull from that cuda-bevfusion repo? Could we possibly clone inside the docker container and then build from that? To avoid cloning during the build process. (Just like patchworks-cpp clones during build this is not ideal)

src/perception/bevfusion/cuda_bevfusion_vendor/CMakeLists.txt
```diff
@@ -0,0 +1,229 @@
...
+            TensorRT_Inc=$ENV{TensorRT_Inc}
+            TensorRT_Lib=$ENV{TensorRT_Lib}
+            ${CMAKE_COMMAND} --build <BINARY_DIR>
+
+        INSTALL_COMMAND ""
+        CONFIGURE_HANDLED_BY_BUILD TRUE
+
+        # Ensure submodules are recursively cloned
+        GIT_SUBMODULES "CUDA-BEVFusion/bevfusion dependencies/stb dependencies/dlpack libraries/3DSparseConvolution libraries/cuOSD"
+    )
```
lucasreljic, line 146: Yea Ideally no cloneing within cmake

src/perception/bevfusion/DEVELOPING.md
```diff
@@ -0,0 +1,169 @@
...
+
+```
+Sensor Inputs
+─────────────
+/multi_camera_sync/multi_image_compressed  ──┐
+(6 cameras, JPEG-compressed, CPU)             │
+                                              ├──► BEVFusionNode (lifecycle composable)
+/lidar_cc/velodyne_points                     │         │
+(configurable LiDAR topic)                 ──┘         │
+                                                         ▼
```
lucasreljic, line 26: Nitpick: fix this offset

src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -449,22 +449,103 @@ multi_camera_sync:
...
+    # Input topics for message filters (remapping does not work for message_filters)
+    input_multi_image_topic: "/multi_camera_sync/multi_image_compressed"  # Topic for synchronized compressed images (sensor_msgs/msg/CompressedImage)
+    input_lidar_topic: "/lidar_cc/velodyne_points"  # Topic for LiDAR point cloud (sensor_msgs/msg/PointCloud2)
+
+    # Directory containing all TensorRT .plan and .onnx engine files
+    model_dir: "/opt/watonomous/models/bevfusion/resnet50int8_trt11"
+    build_dir: "/opt/watonomous/models/bevfusion/resnet50int8_trt11/build"
+
+    # Model precision - can be "fp16" or "int8"
+    precision: "int8"
```
lucasreljic, line 470: if fp16 is put here but hte model_dir is int8 does that cause issues? What does setting int8/fp16 do if the model_dir picks the model? Can this be resolved to either giving the directory of the models then having int8/fp16 selection or have just the model directory and no "precision" setting?

src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -449,22 +449,103 @@ multi_camera_sync:
...
 # =============================================================================
 detection_lifecycle_manager:
   ros__parameters:
     # Manages model inference and attribute enrichment nodes.
-    node_names: ["deep_object_detection_node", "attribute_assigner_node"]
+    node_names: ["bevfusion_node"] # ["deep_object_detection_node", "attribute_assigner_node", "bevfusion_node"]
     autostart: true
-    transition_timeout_s: 30.0
+    autostart_delay_s: 20.0
+    transition_timeout_s: 130.0
```
lucasreljic, line 541: I forget was this needed after the camera_nitros driver changes?

src/perception/perception_bringup/launch/perception.launch.yaml
```diff
@@ -86,41 +86,66 @@ launch:
...
+  #           - name: class_names_path
+  #             value: $(find-pkg-share deep_object_detection)/config/coco_classes.txt
 
   # Load attribute assigner into the shared container
+  # - load_composable_node:
+  #     target: perception_container
+  #     composable_node:
+  #       - pkg: attribute_assigner
+  #         plugin: wato::perception::attribute_assigner::AttributeAssignerNode
+  #         name: attribute_assigner_node
```
lucasreljic, line 117: yea not sure if we want to keep this for traffic light detection for now on just the front camera or leave it commented out. I suggest we have some temporary traffic light detection through this until bevfusion can do traffic light detection

src/perception/bevfusion/bevfusion/test/test_bevfusion_node.cpp
lucasreljic, line 1: Is there a test to see what happens when there is no onnx file or a corrupted onnx file? what is the expected behavoir of that and does it throw a RCLCPP_ERROR msg?

src/perception/perception_bringup/config/perception_bringup.yaml
```diff
@@ -449,22 +449,103 @@ multi_camera_sync:
...
+    # ApproximateTime synchronization
+    sync_queue_size: 5
+    sync_max_time_diff_ms: 200.0
+
+    # QoS
+    qos_subscriber_reliability: "best_effort"
+    qos_subscriber_depth: 10
+    qos_publisher_reliability: "reliable"
+    qos_publisher_durability: "transient_local"
+    qos_publisher_depth: 10
```
lucasreljic, line 530: a depth of 10 is too much for images + lidar, should be like 2 or 3

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
```diff
@@ -0,0 +1,1170 @@
...
+      case 3:
+        r = 0.5f;
+        g = 0.0f;
+        b = 1.0f;
+        break;  // Bus — purple
+      case 4:
+        r = 0.0f;
+        g = 1.0f;
+        b = 1.0f;
+        break;  // Trailer — cyan
```
lucasreljic, line 689: This should really be in a config or some other file, you can add another bevfusion_constants.hpp that has all these constants

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
lucasreljic, line 1: More so a future task, but do you know where the highest processing delay comes from? Looking at the stats from cuda-bevfusion we should be hitting around 20hz for detections, so wondering where that 11hz is from, either upstream (nitros/camera_sync) or within bevfusion node would be good to know

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
```diff
@@ -0,0 +1,1170 @@
...
+    const auto & frame_id = filtered_multi_image_msg->images[i].header.frame_id;
+    cv::Mat bgr = decompressImage(filtered_multi_image_msg->images[i]);
+    if (bgr.empty()) {
+      RCLCPP_WARN_THROTTLE(
+        this->get_logger(), *this->get_clock(), 5000, "Failed to decompress image for frame_id '%s'", frame_id.c_str());
+      return;
+    }
+    rgb_images.emplace_back();
+    cv::cvtColor(bgr, rgb_images.back(), cv::COLOR_BGR2RGB);
+    camera_images.push_back(rgb_images.back().data);
```
lucasreljic, line 237: Can this be cv::cuda:cvtColor? to avoid cpu operations not sure if this can be done later when its already copied to the gpu for inference in the cuda stream

src/perception/bevfusion/bevfusion/src/bevfusion_core.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+  // Camera normalization: resize + mean/std normalization applied to each input image before the backbone
+  ::bevfusion::camera::NormalizationParameter norm;
+  norm.image_width = config_.image_width;
+  norm.image_height = config_.image_height;
+  norm.num_camera = config_.num_cameras;
+  norm.output_width = config_.norm_output_width;
+  norm.output_height = config_.norm_output_height;
+  norm.resize_lim = config_.resize_lim;
+  norm.interpolation = (config_.interpolation == "nearest") ? ::bevfusion::camera::Interpolation::Nearest
```
lucasreljic, line 113: is this the most efficient interpolation?

lucasreljic (changes requested):
I didn't take a thorough look at the core and node itself, but I think from the params there are some changes that can be made to make it more understandable for future development. Let me know if you have further questions about the comments

lucasreljic:
image
should we limit the lidar point cloud to reach the 20fps we want?

lucasreljic:
@AshishA26 Heads up, I was curious about the latency and it seems that jpeg decoding (of the larger 1280x1024) is taking the majority of the callback time, we should figure out a way around this. Either have an option for raw images instead or resize the jpegs before encoding (with nitros).

src/perception/bevfusion/cuda_bevfusion_vendor/CMakeLists.txt
AshishA26: Fixed!

docker/perception.Dockerfile
AshishA26: Yep!

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
AshishA26: Sounds good

src/perception/perception_bringup/config/perception_bringup.yaml
AshishA26: Fixed

src/perception/bevfusion/bevfusion/src/bevfusion_core.cpp
AshishA26: There are 2 types of interpolation - nearest neighbor or bilinear. Nearest is fastest, bilinear is slightly slower but produces better images. I think we can leave it as bilinear and add it to the params file

src/perception/bevfusion/cuda_bevfusion_vendor/CMakeLists.txt
AshishA26: Fixed

src/perception/bevfusion/bevfusion/config/params.yaml
AshishA26: 5 meters below the lidar frame. I'll make it -3.0 instead, which would be more reasonable

src/perception/bevfusion/bevfusion/config/params.yaml
AshishA26: Its the resize multiplier

src/perception/bevfusion/bevfusion/config/params.yaml
AshishA26: So I'll remove the resize_lim from the params. In terms of the code, we actually create a `::bevfusion::camera::NormalizationParameter norm;` which needs the resize limit (`  norm.resize_lim = config_.resize_lim;`).

So instead, I'll automatically make the code calculate the resize limit based on the input and output dimensions

src/perception/bevfusion/DEVELOPING.md
AshishA26: Done

src/perception/bevfusion/bevfusion/config/params.yaml
AshishA26: Actually, it is from lidar frame, but I have to leave it at -5.0, because the model was trained with these specific settings and fails otherwise. CUDA-BEVFusion calculates the 3D voxel grid height like so: `grid_size.z = (max_range.z - min_range.z) / voxel_size.z`, so changing the min range would change the grid size, and thus the number of voxels. 

The lidar.backbone.xyz.onnx) is hard-coded to expect a 3D tensor grid with Z-dimension = 40 apparently, which needs the -5.0 in the min_range.

src/perception/perception_bringup/config/perception_bringup.yaml
AshishA26: Yeah it would cause issues, good catch. Fixed now. Ive made it so that the model is inputted, not precision. Precision is retrieved from the model path

AshishA26:
Fixed jpeg decoding issue
- Brought in an option to use raw images instead. So it now syncs multi_image + lidar, or multi_image_compressed + lidar depending on this option. Needed to keep the compressed option for ros bags.
- Mutlithreaded the jpeg decoding for the compressed images

So far, local testing using compressed image has bumped 13 hz to 20 hz now!

src/perception/bevfusion/bevfusion/test/test_bevfusion_node.cpp
AshishA26: Yes it returns a rclcpp error msg, and I have now added tests for it

lucasreljic (approved):
Changes look good! Great job on getting the 20hz callback speed!

src/perception/perception_bringup/config/perception_bringup.yaml
AshishA26: Looks like its still needed after Friday's testing

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
AshishA26: Seems like Image prep is still highest processing delay.

In general, running bevfusion with the whole autonomy stack is quite heavy

AshishA26:
On car testing has now raised from 11 to 15 Hz! Running the autonomy stack, camera drivers, bevfusion, and more is quite heavy. This rate occurs if only the 6 needed cameras are running (the other cameras can't in order to not cause cuda buffer issues).

A couple things we have to fix further, either in this PR or a separate one:
- We should add a few more profiling time stamps to see which piece of image prep for example is slow
- Use raw images when on-vehicle. The option is there, but does not work currently. Maybe due to multi sync not working for raw images.
- This code (copies images):
```
for (const auto & camera_name : camera_names_) {
  for (const auto & image : multi_image_msg->images) {
    if (image.header.frame_id == camera_name) {
      filtered_multi_image_msg->images.push_back(image);
      break;
    }
  }
}
```
- Limit lidar points
- Potentially try `cv::cuda::cvtColor`
- Using NITROS subscribers
- Resizing upstream before feeding into bevfusion

PR #489, opened 2026-07-10
init commit for 6/20 testing

Matthew0803:
### 📑  Description

This pr is solving 3 problems:
#### Problem 1: Deadman press/release
before: The instant the driver let go of the deadman switch. the steering wheel torque cuts to zero immediately and throttle cuts immediately. If the car was mid-turn, the wheel suddenly go slack/free. When driver press the deadman again to resume, full steering torque and throttle snap back on instantly.

After: Letting go of the deadman fades the steering torque out smoothly over ~600ms instead of an instant cutoff, Pressing the deadman again fades authority back in over ~600ms. Control returns gradually instead of sudden snap.

Note: throttle will still stop immediately, as changing it defeats the purpose of a deadman switch

#### Problem 2: PID (drive mode) → joystick handover
before: driving in PID/manual drive mode and the car is creeping forward slightly at low speed. Switching control to the joystick immediately commands zero throttle. So the car will go from gently creeping forward to abrupt full stop immediately.

after: Handover blends the output from whatever PID was commanding toward the joystick's command over ~600ms. So instead of a sudden stop, the car decelerate smoothly

#### Problem 3: Velocity feedback source (GPS vs CAN)
before: The PID speed controller needs to know the car's actual speed to correct throttle/brake. Right now, due to a wiring bug, it always ends up using GPS-derived speed instead of the CAN wheel-speed sensors.

After: the PID controller reliably locks onto the CAN wheel-speed signal instead, so testing indoors/in GPS-poor areas won't silently degrade the controller anymore.

### ✅  Checklist


src/interfacing/interfacing_bringup/config/interfacing.yaml
```diff
@@ -54,6 +54,13 @@
     toggle_vibration_intensity: 1.0
     toggle_vibration_duration_ms: 150
 
+    # deadman press/release smoothing (steering only — throttle stays instant
+    # for safety). Independent of the OSCC arm/disarm ramp in oscc_interfacing.
+    enable_deadman_ramp: true
```
brianzheng206, line 59: was this tested on the track?

src/interfacing/joystick_interfacing/include/joystick_interfacing/joystick_interfacing.hpp
```diff
@@ -88,6 +88,25 @@ class JoystickNode : public rclcpp_lifecycle::LifecycleNode
   CallbackReturn on_shutdown(const rclcpp_lifecycle::State & state) override;
 
 private:
+  /**
+   * @brief Local smoothing state for the deadman (enable-axis) press/release
+   * transition, applied to steering torque only — independent of the OSCC
+   * arm/disarm ramp in oscc_interfacing. Mirrors that node's DisengageState.
+   *
+   * ENGAGED:     deadman held past the ramp window; steering authority is 1.0.
```
brianzheng206, line 96: is there a reason we use steering authority instead of just straight steering torque? what is authority referring to in this case? 

At disengage, autonomy should instantly have 0 authority (ability to send throttle and steering commands) and should be in handoff mode where its trying to hold the steering wheel position gently, and ramp down the torque so it doesnt flick back to center from road forces

src/interfacing/interfacing_bringup/config/interfacing.yaml
```diff
@@ -185,6 +192,12 @@
       steering: 0.0
       forward: 0.0
 
+    # Handover smoothing: blends from the previous winner's last published
+    # command to the new winner's live command whenever priority handover
+    # changes (e.g. PID creep -> joystick), instead of stepping instantly.
+    enable_handover_ramp: true
+    handover_ramp_ms: 600.0
```
brianzheng206, line 205: handover ramp being 0.6 seconds seems really low..? for reference the average human reaction time is 250ms. i could be misunderstanding what the param is doing, but the time ramping torque down needs to be reactable by the operator

maybe @lucasreljic can give some justice with testing?

brianzheng206: also make sure we arent locking the wheel...

src/interfacing/oscc_mux/src/oscc_mux_node.cpp
```diff
@@ -198,18 +225,66 @@ void OsccMuxNode::mux_callback()
...
+  if (enable_handover_ramp_ && winner != previous_winner_ && previous_winner_ != nullptr) {
+    begin_handover();
+  }
+  previous_winner_ = winner;
+
+  const auto target = winner->last_cmd();
+  roscco_msg::msg::Roscco out;
+  if (in_handover_) {
+    const float a = handover_authority_scale();
+    out.forward = handover_start_cmd_.forward + (target.forward - handover_start_cmd_.forward) * a;
```
brianzheng206, line 250: would this delay our ability to brake by `handover_ramp_ms`?

src/interfacing/pid_control/src/vel_driven_feedforward_pid_node.cpp
```diff
@@ -228,13 +228,12 @@ void VelDrivenFeedforwardPidNode::steering_feedback_callback(const roscco_msg::m
...
-  }
-
-  if (velocity_source_ == VelocitySource::NONE) {
+  // CAN is always authoritative: (re)claim the source on every CAN message,
+  // even if odom_feedback_callback() won a startup race and claimed it first.
+  // Once CAN has published, odom_feedback_callback() permanently ignores odometry.
+  if (velocity_source_ != VelocitySource::CAN) {
     velocity_source_ = VelocitySource::CAN;
-    RCLCPP_INFO(this->get_logger(), "Velocity source locked to CAN (Float64)");
+    RCLCPP_INFO(this->get_logger(), "Velocity source (re)locked to CAN (Float64)");
```
brianzheng206, line 247: is odom even good enough to fallback to? why dont we just wait for CAN?

brianzheng206: also raises transient questions, if CAN dies does it fall back to odom properly

brianzheng206 (commented):
can we update the markdowns as well?

src/interfacing/interfacing_bringup/config/interfacing.yaml
lucasreljic: No it wasn't we need to test this thoroughly before merging

src/interfacing/pid_control/src/vel_driven_feedforward_pid_node.cpp
lucasreljic: Is this prioritizing CAN over Odom? Odom should always be more accurate then CAN velocity as its from the GPS's IMU, the only time is when there is no GPS lock then odom doesn't publish

src/interfacing/interfacing_bringup/config/interfacing.yaml
Matthew0803: We haven't. Will test it next time

src/interfacing/joystick_interfacing/include/joystick_interfacing/joystick_interfacing.hpp
Matthew0803: - authority is the torque scale, that is an unclear naming, will change that

src/interfacing/interfacing_bringup/config/interfacing.yaml
Matthew0803: - handover_ramp_ms is how long the mux crossfades when the winning input source changes
- I dont think wheels are locked as long as we use torque instead of sth like a position servo/angle

lucasreljic:
This has been successfully testing on the car on Aug 1st 2026

src/interfacing/interfacing_bringup/config/interfacing.yaml
```diff
@@ -225,8 +244,10 @@
     enable_brakes: true # Enable brake module on arm (only used when enable_all is false)
     enable_graceful_disarm: true # If true, a manual disarm ramps steering torque to zero before disabling boards (smooth handover). Override/fault/error always disable instantly.
     disarm_ramp_ms: 300.0 # Duration (ms) of the steering torque rampdown on graceful disarm
-    enable_graceful_arm: true # If true, arming fades autonomy authority (steering torque + throttle) in from zero instead of applying full-authority commands instantly. Brake stays at full authority.
-    arm_ramp_ms: 5000.0 # Duration (ms) of the authority ramp-up on graceful arm (10s is a test value; tune down for production)
+    enable_graceful_arm: true # If true, arming fades autonomy authority (steering torque + throttle) in from zero instead of applying full-authority commands instantly.
+    arm_ramp_ms: 5000.0 # Duration (ms) of the authority ramp-up on graceful arm (5s is a test value; tune down for production)
+    brake_ramp_ms: 1000.0 # Duration (ms) of the brake authority ramp-up on graceful arm, so engaging while rolling decelerates smoothly instead of stopping hard.
```
lucasreljic, line 249: This prevents the hard stop when engaging while the car is creeping, 1 second felt like a comfortable fast speed

lucasreljic (approved):
Tested on the car

PR #473, opened 2026-06-30
Documentation + decisions on Nitros subscribers for zero-copy from perception

archangelinux:
### 📑 Description
Updates `DEVELOPING.md` for the BEVFusion package documenting the planned node architecture, sensor inputs, and topic subscriptions. This covers the task of deciding which topics BEVFusion needs to subscribe to for Nitros zero-copy camera data and LiDAR, including the rationale for each choice.

Key decisions documented:
  - 6 Nitros camera topics (`/camera_pano_*/image_rect/nitros`) for zero-copy GPU images
  - `/lidar/all/points_merged` for the pre-merged LiDAR point cloud
  - `/multi_camera_sync/multi_camera_info` for camera intrinsics
  - Camera extrinsics via `tf2_ros::TransformListener` (reads from `/tf_static`)
  - Intended node parameters and build instructions
  
### 📝 Notes for reviewers
 Documentation only, no code changes. Captures the topic research for the Nitros zero-copy subscriber design so the next implementer has a clear reference. Publishers + Parameters table is marked as intended/not yet declared since the node implementation is still in progress.

PR #490, opened 2026-07-10
Implement declareParameters() in BEVFusionNode

archangelinux:
### 📑  Description
  Implements `declareParameters()` in `BEVFusionNode`

  **Files changed:**
  - **`src/bevfusion_node.cpp`** — Implemented `declareParameters()`: declares`model_dir`, `confidence_threshold`, `camera_names`, and the three subscriber topic name parameters. Derives the five TensorRT plan/ONNX file paths from `model_dir` and constructs `BEVFusionCore` with the resulting config. Architecture constants (voxel sizes, BEV grid bounds, etc.) are kept as`BEVFusionInputConfig` defaults and not exposed as ROS params.
  - **`config/params.yaml`** — Added complete parameter file with defaults for all declared parameters: model directory, 6 camera names, topic names, QoS settings, and sync settings.
  - **`test/test_bevfusion_params.cpp`** *(new file)* — Three Catch2 test cases: (1) lifecycle state transitions (unconfigured → inactive → active → inactive → unconfigured), (2) all default parameter values match `params.yaml`, (3) NodeOptions` overrides for `model_dir` and `confidence_threshold` are respected after `configure()`.
  - **`CMakeLists.txt`** — Added `if(BUILD_TESTING)` block wiring up `test_bevfusion_params` via `add_wato_test`.
  - **`package.xml`** — Added `<test_depend>wato_test</test_depend>`.
  - **`DEVELOPING.md`** — Added Testing section with colcon commands.

  ### ✅  Checklist   
  - [x] I listed any breaking changes in the "Notes" section

  ### 🔗  Related Issues / PRs
  Depends on #473 (ashish/bevfusion — BEVFusion node skeleton)

  ### 📝  Notes for reviewers   
  - Build warnings are from existing TODO stubs in `bevfusion_core.cpp`(`initialize()` and `infer()` missing return statements) — not introduced by this PR.
  - `sync_queue_size`, `sync_max_time_diff_ms`, and QoS params are declared in`on_configure()` in Ashish's skeleton, not here, to avoid re-declaring already-registered parameters.
  - Tests run without GPU — `BEVFusionCore` constructor only stores config;`initialize()` (which loads TRT engines) is a TODO and not yet called.
  - This PR adds the first tests for this package

### DIFF
--- a/src/perception/bevfusion/bevfusion/config/params.yaml
+++ b/src/perception/bevfusion/bevfusion/config/params.yaml
@@ -1,3 +1,34 @@
 ---
 bevfusion_node:
   ros__parameters:
+    # Directory containing all TensorRT .plan and .onnx engine files
+    # Place model files at: /opt/watonomous/models/bevfusion/resnet50/
+    model_dir: "/opt/watonomous/models/bevfusion/resnet50"
+
+    # Camera frame IDs — must match the 6 nuScenes-layout cameras on Eve
+    camera_names:
+      - "camera_pano_nn"
+      - "camera_pano_ne"
+      - "camera_pano_nw"
+      - "camera_pano_ss"
+      - "camera_pano_se"
+      - "camera_pano_sw"
+
+    # Detection confidence threshold — boxes below this score are discarded
+    confidence_threshold: 0.3
+
+    # Subscriber topic names
+    topic_multi_image: "/multi_camera_sync/multi_image_compressed"
+    topic_lidar: "/lidar/all/points_merged"
+    topic_camera_info: "/multi_camera_sync/multi_camera_info"
+
+    # ApproximateTime synchronization
+    sync_queue_size: 10
+    sync_max_time_diff_ms: 200.0
+
+    # QoS
+    qos_subscriber_reliability: "best_effort"
+    qos_subscriber_depth: 10
+    qos_publisher_reliability: "reliable"
+    qos_publisher_durability: "transient_local"
+    qos_publisher_depth: 10
--- a/src/perception/bevfusion/bevfusion/config/params.yaml
+++ b/src/perception/bevfusion/bevfusion/config/params.yaml
@@ -17,10 +17,35 @@ bevfusion_node:
     # Detection confidence threshold — boxes below this score are discarded
     confidence_threshold: 0.3
 
-    # Subscriber topic names
-    topic_multi_image: "/multi_camera_sync/multi_image_compressed"
-    topic_lidar: "/lidar/all/points_merged"
-    topic_camera_info: "/multi_camera_sync/multi_camera_info"
+    # Camera input resolution (must match the camera hardware)
+    image_width: 1600
+    image_height: 900
+
+    # Model input resolution after resize
+    norm_output_width: 704
+    norm_output_height: 256
+
+    # LiDAR detection zone in meters [x, y, z]
+    min_range: [-54.0, -54.0, -5.0]
+    max_range: [54.0, 54.0, 3.0]
+
+    # Voxel size in meters [x, y, z]
+    voxel_size: [0.075, 0.075, 0.2]
+
+    # LiDAR point cloud caps (tune for GPU memory vs. coverage tradeoff)
+    max_points_per_voxel: 10
+    max_points: 300000
+    max_voxels: 160000
+
+    # BEV grid bounds [min, max, step] in meters
+    xbound: [-54.0, 54.0, 0.3]
+    ybound: [-54.0, 54.0, 0.3]
+    zbound: [-10.0, 10.0, 20.0]
+    dbound: [1.0, 60.0, 0.5]
+
+    # Bounding box center filter — discard boxes outside this volume [x, y, z]
+    post_center_range_start: [-61.2, -61.2, -10.0]
+    post_center_range_end: [61.2, 61.2, 10.0]
 
     # ApproximateTime synchronization
     sync_queue_size: 10
--- a/src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
+++ b/src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
@@ -57,46 +57,24 @@ BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
 
 void BEVFusionNode::declareParameters()
 {
-  /**
-   * TODO(bevfusion_team)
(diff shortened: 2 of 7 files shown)


src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
```diff
@@ -57,22 +57,38 @@ BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
...
+  this->declare_parameter<std::string>("topic_camera_info", "/multi_camera_sync/multi_camera_info");
+
+  // Derive plan file paths from model_dir using standard CUDA-BEVFusion filenames
   BEVFusionInputConfig config;
+  config.camera_backbone_plan = model_dir + "/camera.backbone.plan";
+  config.camera_vtransform_plan = model_dir + "/camera.vtransform.plan";
+  config.fuser_plan = model_dir + "/fuser.plan";
+  config.head_bbox_plan = model_dir + "/head.bbox.plan";
+  config.lidar_backbone_onnx = model_dir + "/lidar.backbone.onnx";
+  config.confidence_threshold = static_cast<float>(confidence_threshold);
```
AshishA26: Good for now. In the future we might consider putting all params in the BEVFusionInputConfig, into the params.yaml. If you would like to do it in this PR, that is also good with me

src/perception/bevfusion/bevfusion/config/params.yaml
```diff
@@ -1,3 +1,34 @@
...
+      - "camera_pano_se"
+      - "camera_pano_sw"
+
+    # Detection confidence threshold — boxes below this score are discarded
+    confidence_threshold: 0.3
+
+    # Subscriber topic names
+    topic_multi_image: "/multi_camera_sync/multi_image_compressed"
+    topic_lidar: "/lidar/all/points_merged"
+    topic_camera_info: "/multi_camera_sync/multi_camera_info"
```
AshishA26: Topics should use the `remap:` methodology as seen in other nodes and not be named here. Basically in the bevfusion.launch.yaml should have topic names always - you can look at the attribute assigner launch files and nodes as an example. 

Also as a side note: the perception.launch.yaml can override these remappings that you will do, by using `remap:` when it launches the bevfusion node - nothing you should have to do here, just a useful piece of info.

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
```diff
@@ -57,22 +57,38 @@ BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
...
+  // Camera frame IDs, used by on_activate for TF lookups and subscriber topic construction
+  this->declare_parameter<std::vector<std::string>>(
+    "camera_names",
+    std::vector<std::string>{
+      "camera_pano_nn", "camera_pano_ne", "camera_pano_nw", "camera_pano_ss", "camera_pano_se", "camera_pano_sw"});
+
+  // Subscriber topic names; explicit params because message_filters ignores ROS remapping
+  this->declare_parameter<std::string>("topic_multi_image", "/multi_camera_sync/multi_image_compressed");
+  this->declare_parameter<std::string>("topic_lidar", "/lidar/all/points_merged");
+  this->declare_parameter<std::string>("topic_camera_info", "/multi_camera_sync/multi_camera_info");
```
AshishA26: Based on the remap comment I made, these few lines will change and be removed

src/perception/bevfusion/bevfusion/test/test_bevfusion_params.cpp
```diff
@@ -0,0 +1,127 @@
...
+  rclcpp::NodeOptions options;
+  options.append_parameter_override("model_dir", "/custom/test/bevfusion/model");
+  options.append_parameter_override("confidence_threshold", 0.75);
+
+  auto node = std::make_shared<BEVFusionNode>(options);
+  REQUIRE(node->configure().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE);
+
+  REQUIRE(node->get_parameter("model_dir").as_string() == "/custom/test/bevfusion/model");
+  REQUIRE_THAT(node->get_parameter("confidence_threshold").as_double(), Catch::Matchers::WithinRel(0.75));
+}
```
AshishA26, line 73: Thank you for adding tests! Would it be possible for you to change the structure to be using `#include <wato_test/wato_test.hpp>` instead? There is a dedicated package that is in the monorepo and `src/wato_test/README.md` will have examples, as well as other nodes will have good examples on how to use it.

Also not sure entirely, but I don't think we really should test the params itself usually, unless you are making a mock a params file or something? For example, if one were to change a parameter for qos to be from best_effort to reliable, they would also have to change the tests. These tests are basically forcing a certain set of params, unless there is another way to do it. I would recommend seeing how other nodes go about doing tests (ctrl+shift+F and search for places that do `wato_test` or `wato::test`

AshishA26 (changes requested):
Needs some changes, but great work overall!

src/perception/bevfusion/bevfusion/config/params.yaml
archangelinux: Done, removed the three topic params from params.yaml and added them as launch args with defaults and `remap:` entries in bevfusion.launch.yaml, following the attribute assigner pattern

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
archangelinux: Removed from declareParameters(). (Topic name parameters for message_filters will be declared inside on_configure() when subscriptions are set up, and there's already a TODO there about this.)

src/perception/bevfusion/bevfusion/src/bevfusion_node.cpp
archangelinux: sounds good - threw it all into params.yaml in this PR

src/perception/bevfusion/bevfusion/test/test_bevfusion_params.cpp
archangelinux: Switched to #include <wato_test/wato_test.hpp> and followed the pattern from test_tracking.cpp (inline rclcpp::init/shutdown, no custom RAII guard). Removed the default value test (Test 2), and kept just the ones that don't test specific values(lifecycle state machine test and the NodeOptions override test)

AshishA26 (approved):
Looks good!

PR #492, opened 2026-07-22
Implement bevfusion core initialize

archangelinux:
### 📑 Description
Implements `BEVFusionCore::initialize()`, which translates `BEVFusionInputConfig` into the CUDA-BEVFusion vendor library's parameter structs and calls `create_core()` to deserialize all TensorRT engine plans into GPU-executable engines. Also adds the destructor (CUDA stream cleanup) and `isInitialized()` guard flag. 
`infer()` and `updateCalibration()` remain as documented TODOs for follow-up PRs.

### 📹 (Optional) Video Demo of Changes
N/A — no runtime output until `infer()` and `on_configure()` are implemented.

### ✅ Checklist
- [x] I listed any breaking changes in the "Notes" section

### 🔗 Related Issues / PRs
First PR in series of PRs for bevfusion-core
Merging into ashish/bevfusion

### 📝 Notes for reviewers
- `initialize()` must be called explicitly after construction (two-phase init) — the constructor is intentionally cheap and fallible work is deferred to `initialize()`
- `libcustom_layernorm.so` is loaded via `dlopen` before `create_core()` — TRT cannot deserialize `head.bbox.plan` without this plugin registered first
- `isInitialized()` returns false until `initialize()` succeeds; callers (`on_activate`, syncedCallback`) should guard on this before calling `infer()`
- The fast unit test (`[core][fast]`) requires no GPU — only verifies the guard flag at construction

### DIFF
--- a/src/perception/bevfusion/bevfusion/CMakeLists.txt
+++ b/src/perception/bevfusion/bevfusion/CMakeLists.txt
@@ -180,6 +180,11 @@ if(BUILD_TESTING)
     LIBRARIES
       ${PROJECT_NAME}_node
   )
+  add_wato_test(test_bevfusion_core
+    test/test_bevfusion_core.cpp
+    LIBRARIES
+      ${PROJECT_NAME}_wato_core
+  )
 endif()
 
 ament_package()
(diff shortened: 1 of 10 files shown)


AshishA26 (approved):
Made a quick change - I realized we should add precision as an input param, and then added a comment stating that nvtype comes from the cuda-bevfusion repo.

Looks good! Approved

PR #508, opened 2026-08-06
Elastic Trajectory Planner

jasonkdeng:
### 📑  Description

Elastic Trajectory Planner implemented. Tests succeed. However, I am having trouble with WSL running world modeling.

### ✅  Checklist

### 📝  Notes for reviewers

This PR is not complete, and requires assistance with debugging. Currently, running world modelling is crashing this.


jasonkdeng:
I suspect the issue is with RAM, because there are no immediate errors when running world modeling, and instead, WSL disconnects after about 10 seconds.

PR #513, opened 2026-08-13
Add arg to disable live cameras

AshishA26:
### 📑  Description

Launching the perception_bringup with perception.launch.yaml use to fail as the camera drivers don't exist anywhere except when on the car directly. This PR adds a flag preventing loading in physical cameras.

### ✅  Checklist

### 🔗  Related Issues / PRs
Fixes #123
Depends on #456


lucasreljic (approved):
Looks good! Thanks for making default true so this should not break deploy container

